PR #13, opened 2024-03-15
Fix code gen task where cargo.toml is changed

youyuanwu:
From https://github.com/microsoft/windows-rs/issues/2931
it is possible to re-generate the code without duplicated feature entries.
Also enables build the code after regeneration in ci.


rsajwani:
looks good to me

rsajwani (approved):

PR #16, opened 2024-03-26
Fix all cargo fmt and clippy errors and enable check in ci

youyuanwu:


crates/samples/echomain-stateful/src/app.rs
```diff
@@ -287,8 +287,8 @@ pub struct AppInstance {
     hostname_: HSTRING,
     tx_: Cell<Option<Sender<()>>>, // hack to use this mutably
     th_: Cell<Option<JoinHandle<Result<(), Error>>>>,
-    role_: Cell<mssf_com::FABRIC_REPLICA_ROLE>,
-    replicator_: Cell<Option<IFabricReplicator>>,
+    // role_: Cell<mssf_com::FABRIC_REPLICA_ROLE>,
```
rsajwani, line 290: nit: why not delete  it instead of comment..

rsajwani (approved):

crates/samples/echomain-stateful/src/app.rs
youyuanwu: It will be brought back for any meaningful example.

PR #18, opened 2024-04-16
Refactor Fabric Sender and Receiver for client

youyuanwu:
Refactored Sender Receiver to unwrap itself rather than repeating unwraps in generated code.

### DIFF
--- a/crates/libs/core/src/client/svc.rs
+++ b/crates/libs/core/src/client/svc.rs
@@ -13,23 +13,21 @@ impl IFabricServiceManagementClient6Wrap {
         description: &::mssf_com::Microsoft::ServiceFabric::FABRIC_SERVICE_DESCRIPTION,
         timeoutMilliseconds: u32,
     ) -> crate::sync::FabricReceiver<::windows_core::Result<()>> {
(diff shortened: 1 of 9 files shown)


crates/libs/core/src/client/svc.rs
```diff
@@ -52,78 +48,72 @@ impl IFabricServiceManagementClient6Wrap {
             )
         };
         if ctx.is_err() {
-            let (tx2, rx2) = tokio::sync::oneshot::channel();
-            tx2.send(Err(ctx.err().unwrap())).expect("fail to send tx2");
-            crate::sync::FabricReceiver::new(rx2)
+            let (tx2, rx2) = crate::sync::oneshot_channel();
+            tx2.send(Err(ctx.err().unwrap()));
```
rsajwani, line 52: why remove expect here?

rsajwani (approved):

crates/libs/core/src/client/svc.rs
youyuanwu: send  internally will unwrap.

PR #24, opened 2024-04-22
Various Improvements

youyuanwu:
* Add type in the config parameter get load.
* Refactor iterator implementation for FabricClient.
* Refactor the use of raw pointer in mgmt client.


crates/libs/core/src/iter.rs
```diff
@@ -1,4 +1,8 @@
 // iterator implementation
+// Iter infrastruction to convert Fabric raw list into rust safe wrappers.
```
qbradley, line 2:
```suggestion
// Iter infrastructure to convert Fabric raw list into rust safe wrappers.
```

qbradley (approved):

PR #25, opened 2024-04-24
Integrate with config-rs

youyuanwu:
Added Fabric Config Package as a source provider to config-rs.
Added test in the echo app to load the config and deserialize to struct.
Config-rs can load the config. We can utilize config-rs to get config in a standard way or deserialize it in to structs.
See https://github.com/mehcode/config-rs

### DIFF
--- a/crates/samples/echo2/src/main.rs
+++ b/crates/samples/echo2/src/main.rs
@@ -4,12 +4,18 @@
 // ------------------------------------------------------------
 
 use log::info;
-use mssf_core::runtime::{
-    executor::{DefaultExecutor, Executor},
-    ActivationContext, Runtime,
+use mssf_core::{
+    conf::{Config, FabricConfigSource},
+    runtime::{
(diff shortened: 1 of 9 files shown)


crates/samples/echo2/src/main.rs
```diff
@@ -32,6 +38,7 @@ fn main() -> windows::core::Result<()> {
     Ok(())
 }
 
+// validates the configs in the config package have the right values.
 fn validate_configs(actctx: &ActivationContext) {
```
rsajwani, line 42: Will this fail if there is no confi? seems lke is the case

rsajwani (approved):

crates/samples/echo2/src/main.rs
youyuanwu: The app will crash if config is not found or fail to load. This is expected behavior and we rely this in CI to make sure the config is actually working. If the app crashes , the ci should fail.

PR #26, opened 2024-04-24
Allow mssf-core user to not depend on windows crate, and merge 2 test apps into 1

youyuanwu:
We have 2 sample apps for stateless service, and each tests some part of the mssf-core and com wrapping code. We only run echomain app in the CI, so some of the test paths were not executed.
Merging 2 apps in a single one so that CI can cover all the test paths.

Also re-exported some times from windows crate in mssf-core, so that users of this crate does not have to explicitly depend of windows crate.


crates/samples/echomain/src/app.rs
```diff
@@ -3,183 +3,116 @@
...
-        initializationdatalength: u32,
-        initializationdata: *const u8,
-        partitionid: &::windows::core::GUID,
+        servicetypename: &HSTRING,
+        servicename: &HSTRING,
+        initializationdata: &[u8],
+        partitionid: &mssf_core::GUID,
         instanceid: i64,
-    ) -> ::windows::core::Result<IFabricStatelessServiceInstance> {
-        let mut init_data: String = "".to_string();
```
rsajwani, line 66: what is the use-case of initialization-data ? is this code moved under safe wrapper or are we completely removing it.

youyuanwu: Old code is using raw ptr and buffer length:
```rs
 initializationdatalength: u32,
        initializationdata: *const u8,
```
new code is using a slice that has the buffer ref and length inside.
```rs
initializationdata: &[u8],
```
The functionality is the same, but wrapped.

rsajwani (approved):

PR #30, opened 2024-04-29
Migrate from Log to Tracing

youyuanwu:
Migrate to use tracing crate:
https://github.com/tokio-rs/tracing/tree/master


rsajwani:
Didn't get enough time to read. but is there some default config for Tracing which configure default output to console etc .. What's the benefit of using tracing here?

youyuanwu:
> Didn't get enough time to read. but is there some default config for Tracing which configure default output to console etc .. What's the benefit of using tracing here?

We have migrated to tracing internally already. Tracing is a frontend to format and cache the traces, app needs to setup subscriber at the beginning of the app bootup. So tracing in this lib and tracing write to disk/console are separate.

rsajwani (approved):

PR #33, opened 2024-05-06
Correctly validate pointers are non-null

jrahman:
Rust 1.78 will now validate preconditions to unsafe standard library functions. This includes std::slice::from_raw_parts which validates the pointer is:
1) Non-null
2) Correctly aligned

In this case, initializationdata can be null, so we need to correctly address that by returning an empty slice explicitly rather than calling from_raw_parts() will a nullptr.


youyuanwu (approved):
This should fix the issue.
You can try upgrade the rs version in this repo as well.
And you need to release/publish new crates(including update lib versions) for downstream to consume.

qbradley (approved):

PR #37, opened 2024-06-12
Add Tokio and config-rs features

cgettys-microsoft:
As discussed offline, I have some oddball integration scenarios where much of the SF runtime interaction is already handled by existing application logic written in other languages, but we'd like to move parts of the system into Rust.

At least for now, my project is not making use of any of the functionality in mssf-core that actually necessitates use of Tokio, nor are we currently using the config_rs crate. At the same time, mssf_com is a bit too low level to be ideal for us - we have no desire to duplicate / rewrite existing Rust wrappers such as:
* mssf_core::runtime::config types
* mssf_core::runtime::ActivationContext

To enable minimizing dependencies and compile times for those of use who don't need the entire runtime at present, this PR adds two default-on features (naming suggestions of course welcome) - one for config-rs, one for tokio (tokio feature also requires ctrlc as that's small and part of the Executor, and I'm assuming most native-Rust applications that are ok with tokio will want that too).

It also moves some of the code form runtimes/mod.rs to its own file to make the conditional complication a bit less messy, though it could be cleaner.

Additionally, one of the echomain examples is adjusted to demonstrate that the config-rs feature is optional. I h aven't shown the Tokio one. it does compile appropriately, but it would require a C# or C++ demo application embedding a Rust dll or static library containing this code - all the examples use callbacks or service factory types that I can see, and those APIs require Tokio presently.

### DIFF
--- a/crates/libs/core/src/runtime/mod.rs
+++ b/crates/libs/core/src/runtime/mod.rs
@@ -5,17 +5,15 @@
 
 use mssf_com::{
     FabricCommon::FabricRuntime::{
-            FabricCreateRuntime, FabricGetActivationContext, IFabricCodePackageActivationContext,
-            IFabricRuntime,
-        },
+        FabricCreateRuntime, FabricGetActivationContext, IFabricCodePackageActivationContext,
+        IFabricRuntime,
+    },
     FABRIC_ENDPOINT_RESOURCE_DESCRIPTION,
 };
 use windows_core::{Error, Interface, HSTRING, PCWSTR};
-    
+
 #[cfg(feature = "tokio")]
-use mssf_com::FabricCommon::{
-        IFabricAsyncOperationCallback, IFabricAsyncOperationContext,
-    };
+use mssf_com::FabricCommon::{IFabricAsyncOperationCallback, IFabricAsyncOperationContext};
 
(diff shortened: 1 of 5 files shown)


crates/libs/core/src/runtime/mod.rs
```diff
@@ -4,35 +4,43 @@
...
+
+#[cfg(feature = "tokio_async")]
 mod bridge;
 pub mod config;
 pub mod error;
+#[cfg(feature = "tokio_async")]
 pub mod executor;
+#[cfg(feature = "tokio_async")]
 pub mod node_context;
+#[cfg(feature = "tokio_async")]
```
youyuanwu, line 31: Could be simplified by using cfg-if crate: https://crates.io/crates/cfg-if

youyuanwu (approved):

crates/libs/core/src/runtime/mod.rs
cgettys-microsoft: Isn't that mostly for else-if type chainign?

Using a mod in this file might help to wrap them all up together

youyuanwu: You can use a mod and reexport it as well.

PR #40, opened 2024-06-18
Safe wrapping for MustCatchup field in FABRIC_REPLICA_INFORMATION_EX1

### LINKED ISSUE #39: PrimaryReplictorProxy crashes on build_replica()
Build replica crashed in SF runtime.
```rs
async fn build_replica(&self, replica: &ReplicaInfo) -> ::windows_core::Result<()>;
```
The problem is that the MustCatchup field is not correctly propagated from proxy layer into BeginBuildReplica c function.
This affects only apps using PrimaryReplicatorProxy.

youyuanwu:
Fixes #39

### DIFF
--- a/crates/libs/core/src/runtime/stateful_types.rs
+++ b/crates/libs/core/src/runtime/stateful_types.rs
@@ -5,9 +5,11 @@
 
 // stateful_types contains type wrappers for sf stateful raw types
 
+use std::ffi::c_void;
+
 use mssf_com::{
-    FABRIC_EPOCH, FABRIC_REPLICA_INFORMATION, FABRIC_REPLICA_OPEN_MODE,
-    FABRIC_REPLICA_OPEN_MODE_EXISTING, FABRIC_REPLICA_OPEN_MODE_INVALID,
+    FABRIC_EPOCH, FABRIC_REPLICA_INFORMATION, FABRIC_REPLICA_INFORMATION_EX1,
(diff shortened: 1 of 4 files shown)


crates/libs/core/src/runtime/stateful_types.rs
```diff
@@ -129,88 +133,134 @@ impl From<FABRIC_REPLICA_STATUS> for ReplicaStatus {
...
             CatchUpCapability: r.CatchUpCapability,
+            MustCatchUp: must_catchup,
         }
     }
 }
 
 impl ReplicaInfo {
-    pub fn get_raw(&self) -> FABRIC_REPLICA_INFORMATION {
-        FABRIC_REPLICA_INFORMATION {
+    // The parts has the same lifetime as self.
```
jrahman, line 247: nit
```suggestion
    // The parts have the same lifetime as self.
```

crates/libs/core/src/runtime/stateful_types.rs
```diff
@@ -129,88 +133,134 @@ impl From<FABRIC_REPLICA_STATUS> for ReplicaStatus {
...
+    // Caller needs to stitch the parts together.
```
jrahman, line 248: nit: What does "stitch the parts together" mean here? From background conversations, FABRIC_REPLICA_INFORMATION::Reserved needs to point at FABRIC_REPLICA_INFORMATION_EX1, but would not be immediately obvious to a fresh reader

jrahman (approved):

rsajwani (approved):

crates/libs/core/src/runtime/stateful_types.rs
youyuanwu: Added the details.

PR #43, opened 2024-06-28
Upgrade windows-rs to version 0.57

youyuanwu:
windows-rs version 0.57 has various improvements for generated com bindings and the code size reduces.

We have been using win-bindgen with config=package, which is not publicly supported, and we are advised to switch to config=implement that is supported external to windows-rs. See https://github.com/microsoft/windows-rs/issues/3100 for details.

The generated code are reorganized due to the generation config mode, and the mod structures are remapped so that there is no unwanted nesting of mods, for example, FabricRuntime is no longer inside FabricCommon but they are siblings now. The generation of the code and remapping of the mods are still hacky, and we still need https://github.com/Azure/service-fabric-metadata/issues/1 to improve this.

Previously all mods are generated once, but now, each mod like FabricCommon and FabricRuntime is generated separately from the winmd file per pass. The generation speed maybe slowed but in practice it is not observable.

This upgrade will cause break in down stream because of the mod layout change. But it is easy to fix: just change the imports(use) with different path.


crates/tools/fabric-gen/src/gen.rs
```diff
@@ -62,16 +62,42 @@ impl TypeEntry {
...
+                    // TODO: this might not be right?
+                    &TypeName("Windows.Win32.Foundation", "BOOLEAN") => (
+                        "BOOLEAN".to_string(),
+                        "::windows_core".to_string(),
+                        empty_str,
+                    ),
+                    _ => panic!("typename not supported: {} {}", n.namespace(), n.name()),
+                }
             }
+            // Type::HRESULT => (
```
jrahman, line 91: Remove commented code?

jrahman (approved):
Didn't have a chance to review each file in detail, but overall seems fine

crates/libs/core/src/client/mod.rs
```diff
@@ -56,21 +56,21 @@ impl FabricClient {
...
     // Get the client for managing Fabric Properties in Naming Service
-    pub fn get_property_manager(&self) -> PropertyManagementClient {
-        PropertyManagementClient {
-            _com: self.com_property_client.clone(),
-            _gen_wrap: IFabricPropertyManagementClient2Wrap::from_com(
-                self.com_property_client.clone(),
-            ),
-        }
-    }
+    // pub fn get_property_manager(&self) -> PropertyManagementClient {
```
rsajwani, line 66: commented code??

rsajwani (approved):

rsajwani (approved):

crates/tools/fabric-gen/src/gen.rs
youyuanwu: file removed from the generator removal PR.

PR #49, opened 2024-07-11
Impl fabric error conversion

youyuanwu:
SF uses FARBIC_ERROR_CODE and some windows common HRESULT to pass errors between COM and app boundaries.
This PR provides conversion infra for FARBIC_ERROR_CODE and provides some common HRESULT values in mssf-core crate, so the users does not need to depend on windows crates to get error constants.

### DIFF
--- a/crates/libs/core/src/error/mod.rs
+++ b/crates/libs/core/src/error/mod.rs
@@ -1,0 +1,107 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


crates/libs/core/src/error/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+impl From<FabricError> for HRESULT {
+    fn from(value: FabricError) -> Self {
+        value.0
+    }
+}
+
+// SF uses win32 hresult code together with the fabric error code.
```
jrahman, line 48: nit: // -> /// for Rust doc

youyuanwu: fixed.

jrahman (approved):

PR #53, opened 2024-07-18
Refactor rust to com proxy wrapper

youyuanwu:
Refactors the wrapper for rust code calling SF com Begin and End api.
The duplicated code for each api wrapping is eliminated, and the wrapping is easier and more readable. The lines of code is reduced significantly.

### DIFF
--- a/crates/libs/core/src/client/query_client.rs
+++ b/crates/libs/core/src/client/query_client.rs
@@ -18,7 +18,7 @@ use mssf_com::{
 
 use crate::{
     strings::get_pcwstr_from_opt,
-    sync::{self, FabricReceiver},
+    sync::{fabric_begin_end_proxy, FabricReceiver},
     types::{
         NodeList, NodeQueryDescription, ServicePartitionList, ServicePartitionQueryDescription,
     },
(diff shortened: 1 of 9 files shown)


crates/libs/core/src/client/query_client.rs
```diff
@@ -37,48 +37,29 @@ impl QueryClient {
...
-        } else {
-            rx
-        }
+        let com1 = &self.com;
+        let com2 = self.com.clone();
+        fabric_begin_end_proxy(
+            move |callback| unsafe {
+                com1.BeginGetNodeList(queryDescription, timeoutMilliseconds, callback)
+            },
+            move |ctx| unsafe { com2.EndGetNodeList2(ctx) },
```
rsajwani, line 46: how will I do the custom error handling in this case?

rsajwani: can I still get rx object here

youyuanwu: The semantic is unchanged as before.
The error of the begin and end com function will be propagated to the rx receiver in the content and returned to the user.
See the FabricReceiver content type is wrapped in "Result".

rsajwani (approved):

Issue #46, opened 2024-06-28
Feature: Handle Cancel for IFabricAsyncOpertionContext


youyuanwu:
In SF com api, cancellation is handled by IFabricAsyncOperationContext:
https://github.com/Azure/service-fabric-rs/blob/acb9afc3fbf211803874480714e3a3f9f001efda/crates/libs/com/src/ServiceFabric/FabricCommon.rs#L216

The rust style wrapping of various SF com apis ignores this cancellation, and informs SF that the operation is not cancellable.
Various Abort signals may also relate to cancellation.

In csharp api looks like this:
```csharp
public System.Threading.Tasks.Task<System.Fabric.Query.NodeList> GetNodeListAsync (string nodeNameFilter, 
System.Fabric.Query.NodeStatusFilter nodeStatusFilter, string continuationToken, TimeSpan timeout, 
System.Threading.CancellationToken cancellationToken);
```
Where `CancellationToken cancellationToken` can trigger SF c++ to cancel the operation. In SF -> csharp code as well, CancellationToken is passed to tell csharp code to cancel.

For Rust, a similar api can be modeled follwoing csharp:
Current Rust -> SF api:
```rs
    pub async fn get_node_list(
        &self,
        desc: &NodeQueryDescription,
        timeout: Duration,
    ) -> windows_core::Result<NodeList>
```
Proposed client api: Rust code can call token cancel to cancel SF operation.
```rs
    pub async fn get_node_list(
        &self,
        desc: &NodeQueryDescription,
        timeout: Duration,
        cancelation_token:  tokio_util::sync::CancellationToken
    ) -> windows_core::Result<NodeList>
```
Current sf->rust api
```rs
    async fn open(
        &self,
        openmode: OpenMode,
        partition: &StatefulServicePartition,
    ) -> windows::core::Result<impl PrimaryReplicator>;
```
Proposed sf -> rust api
```rs
    async fn open(
        &self,
        openmode: OpenMode,
        partition: &StatefulServicePartition,
        cancelation_token:  tokio_util::sync::CancellationToken
    ) -> windows::core::Result<impl PrimaryReplicator>;
```
When writing rust app, one can check if sf cancels the operation from cancelation_token.

youyuanwu:
Completed in #65

PR #54, opened 2024-07-18
Refactor bridge implementation for COM to rust

youyuanwu:
Refactored the common code for making bridge for COM api calling async rust code.
The result code has better readability and the number of lines of wrapping code is reduced. In future the wrapping for new COM to rust api should be much eaiser.
This also paves the way to implement a consistent wrapping for cancelation in #46 .


rsajwani (approved):

crates/libs/core/src/sync/bridge.rs
```diff
@@ -0,0 +1,61 @@
...
+use super::bridge_context::BridgeContext;
+
+/// Bridge for SF Begin com api.
+/// This launches the future task on rt, and triggers the callback when
+/// the future completes.
+/// The returned context must be passed up to the Begin api's return.
+/// The End api must use fabric_end_bridge() to retrieve the async result from
+/// the returned context.
+/// Remarks:
+/// This api in a sense is unsafe because the develper needs to ensure the following:
```
jrahman, line 24: What are the consequences of violating this invariant?

jrahman (approved):
Great cleanup!

crates/libs/core/src/sync/bridge.rs
youyuanwu: Access violation or memory corruption. I may add some debug checks in future.

PR #55, opened 2024-07-19
[FabricClient] implement restart replica api

youyuanwu:
Implemented restart replica fabric client api for testing stateful service replica restart. This API will be used to test replica reopen workflow implementation.
Tested for the restarting of the example stateful service.

### DIFF
--- a/crates/libs/core/src/client/svc_mgmt_client.rs
+++ b/crates/libs/core/src/client/svc_mgmt_client.rs
@@ -11,18 +11,20 @@ use mssf_com::{
         FABRIC_PARTITION_KEY_TYPE, FABRIC_PARTITION_KEY_TYPE_INT64,
         FABRIC_PARTITION_KEY_TYPE_INVALID, FABRIC_PARTITION_KEY_TYPE_NONE,
         FABRIC_PARTITION_KEY_TYPE_STRING, FABRIC_RESOLVED_SERVICE_ENDPOINT,
-        FABRIC_SERVICE_ENDPOINT_ROLE, FABRIC_SERVICE_PARTITION_KIND,
-        FABRIC_SERVICE_PARTITION_KIND_INT64_RANGE, FABRIC_SERVICE_PARTITION_KIND_INVALID,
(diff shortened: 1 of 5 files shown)


crates/libs/core/src/client/svc_mgmt_client.rs
```diff
@@ -87,6 +108,20 @@ impl ServiceManagementClient {
         let res = ResolvedServicePartition::from_com(com);
         Ok(res)
     }
+
+    /// Simulates a service replica failure by restarting a persisted service replica,
+    /// closing the replica, and then reopening it. Use this to test your service for problems
+    /// along the replica reopen path. This helps simulate the report fault temporary path through client APIs.
+    /// This is only valid for replicas that belong to stateful persisted services.
+    pub async fn restart_replica(
```
rsajwani, line 116: should we name it as restart_stateful_replica?

qbradley (approved):

crates/samples/echomain-stateful2/src/test.rs
```diff
@@ -60,14 +60,23 @@ async fn test_partition_info() {
...
+    assert_ne!(stateful_replica.node_name, HSTRING::new());
+
+    let desc = RestartReplicaDescription {
+        node_name: stateful_replica.node_name.clone(),
+        partition_id: single.id,
+        replica_or_instance_id: stateful_replica.replica_id,
+    };
+    let mgmt = fc.get_service_manager();
+    mgmt.restart_replica(&desc, timeout).await.unwrap();
 }
```
rsajwani, line 82: should we have assert after restart replica

rsajwani (approved):

crates/libs/core/src/client/svc_mgmt_client.rs
youyuanwu: I prefer keep the same name as the COM api.

crates/samples/echomain-stateful2/src/test.rs
youyuanwu: Once we have more apis ready I will add checks. The checks here needed is for query service health and status, which is not implemented yet.

PR #56, opened 2024-07-20
Stateful partition report load api

youyuanwu:
Add api for report load from replica to SF. SF Load balancer will balancing the services according to the metric reported.
Tested in the stateful echo test app and reported metrics showed up in SFE. Once metrics-get FabricClient API is implemeneted, more tests can be written.


crates/libs/core/src/runtime/stateful.rs
```diff
@@ -43,20 +43,28 @@ pub trait LocalStatefulServiceReplica: Send + Sync + 'static {
...
 }
 
 impl StatefulServicePartition {
     pub fn get_com(&self) -> &IFabricStatefulServicePartition {
-        &self._com_impl
+        &self.com_impl
+    }
+
+    /// Reports load for the current replica in the partition.
+    pub fn report_load(&self, metrics: &[LoadMetric]) -> crate::Result<()> {
```
jrahman, line 57: To confirm my understanding, the application itself periodically calls `report_load()` itself to report load information. SF itself does not call this as a callback, but instead the application is responsible for calling this on some cadence

jrahman (approved):

crates/libs/core/src/runtime/stateful.rs
youyuanwu: Yes. Application is responsible for calling this API. If app does not report load, the default load in the app/service manifest will be used by SF. App reported load will replace the default load, and the latest reported load will be used by SF for balancing decisions.

PR #58, opened 2024-08-09
Initial support for cancellation

youyuanwu:
Implement #46 
All SF rust api should have a new parameter `cancellation_token: CancellationToken`. CancellationToken is taken directly from tokio_util crate, as I don't see an advantage to do a wrapping of it or creating our own implementation of cancellation token. SF treats cancellation as an advisory signal, and in user application(rust api), user can ignore the cancel signal, but it might stall SF until the operation is finished.

In this PR there is only 1 proxy api and 1 bridge api has been modified to have this cancellation token, and I will create another PR to change all apis.
The implementation requires new BridgeContext, oneshot Sender/Receiver, and fabric_begin_end_proxy. The original implementations are not changed in this PR but will be removed/replaced in the next PR.

### DIFF
--- a/crates/libs/core/src/sync/cancel.rs
+++ b/crates/libs/core/src/sync/cancel.rs
@@ -18,6 +18,7 @@ use windows_core::{implement, AsImpl};
 
 use crate::{error::FabricErrorCode, runtime::executor::Executor};
 
+/// Async operation context for bridging rust code into SF COM api that supports cancellation.
 #[implement(IFabricAsyncOperationContext)]
 pub struct BridgeContext3<T>
 where
@@ -44,13 +45,23 @@ where
         }
     }
 
+    /// Creates the context from callback, and returns a cancellation token that
+    /// can be used in rust code, and the cancellation token is hooked into self,
+    /// where Cancel() api cancels the operation.
     pub fn make(callback: Option<&IFabricAsyncOperationCallback>) -> (Self, CancellationToken) {
         let token = CancellationToken::new();
         let ctx = Self::new(callback.unwrap().clone(), token.clone());
         (ctx, token)
     }
 
-    // executes the future in background
+    /// Spawns the future on rt.
+    /// Returns a context that can be returned to SF runtime.
+    /// This is intended to be used in SF Begin COM api, where
+    /// rust code is spawned in background and the context is returned
+    /// to caller.
+    /// This api is in some sense unsafe, because the developer needs to ensure
+    /// the following:
+    /// * return type of the future needs to match SF COM api end return type.
     pub fn spawn<F>(
         self,
         rt: &impl Executor,
@@ -72,6 +83,13 @@ where
         Ok(self_cp2)
     }
 
+    /// Get the result from the context from the SF End COM api.
+    /// This api is in some sense unsafe, because the developer needs to ensure
+    /// the following:
+    /// * context impl type is `BridgeContext3`, and the T matches the SF end api
+    /// return type.
+    /// Note that if T is of Result<ICOM> type, the current function return type is
+    /// Result<Result<ICOM>>, so unwrap is needed.
     pub fn result(context: Option<&IFabricAsyncOperationContext>) -> crate::Result<T> {
         let self_impl: &BridgeContext3<T> = unsafe { context.unwrap().as_impl() };
         self_impl.consume_content()
@@ -137,6 +155,8 @@ impl<T> IFabricAsyncOperationContext_Impl for BridgeContext3<T> {
 pub struct FabricReceiver2<T> {
     rx: tokio::sync::oneshot::Receiver<T>,
     token: Option<CancellationToken>,
+    // saved ctx from SF Begin COM api for cancalling.
+    ctx: Option<IFabricAsyncOperationContext>,
 }
 
 impl<T> FabricReceiver2<T> {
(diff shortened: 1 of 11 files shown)


rsajwani (changes requested):

crates/libs/core/src/sync/cancel.rs
```diff
@@ -0,0 +1,729 @@
...
+            self_impl.set_content(ok);
+            self_impl.set_complete();
+            let cb = self_impl.Callback().unwrap();
+            unsafe { cb.Invoke(&self_cp) };
+        });
+        Ok(self_cp2)
+    }
+
+    /// Get the result from the context from the SF End COM api.
+    /// This api is in some sense unsafe, because the developer needs to ensure
```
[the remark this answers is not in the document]
youyuanwu: The existing code without cancellation has the same behaviour.
Developer writing this crate should be careful of this and test the apis using this. In future I might add some debug_asserts but that requires more work.

crates/libs/core/src/sync/cancel.rs
```diff
@@ -0,0 +1,729 @@
...
+                    // this is cancel safe so we can poll it once and discard
+                    let fu = t.cancelled();
+                    let inner = std::pin::pin!(fu).poll(_cx);
+                    match inner {
+                        Poll::Ready(_) => {
+                            // operation cancelled
+                            if let Err(e) = self.cancel_inner_ctx() {
+                                return Poll::Ready(Err(e));
+                            }
+                            Poll::Ready(Err(FabricErrorCode::OperationCanceled.into()))
```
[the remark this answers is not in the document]
youyuanwu: It means that the Future has finished but the Result is an error. User awaits for the future will be unblocked, but the result the user gets is that Operation is cancelled.

crates/libs/core/src/sync/cancel.rs
```diff
@@ -0,0 +1,729 @@
...
+            token,
+        }
+    }
+
+    /// Creates the context from callback, and returns a cancellation token that
+    /// can be used in rust code, and the cancellation token is hooked into self,
+    /// where Cancel() api cancels the operation.
+    pub fn make(callback: Option<&IFabricAsyncOperationCallback>) -> (Self, CancellationToken) {
+        let token = CancellationToken::new();
+        let ctx = Self::new(callback.unwrap().clone(), token.clone());
```
jrahman, line 53: If this code always calls ::unwrap() on `callback`, why not change the signature to `make(callback: &IFabricAsyncOperationCallback)`?

crates/libs/core/src/sync/cancel.rs
```diff
@@ -0,0 +1,729 @@
...
+    /// return type.
+    /// Note that if T is of Result<ICOM> type, the current function return type is
+    /// Result<Result<ICOM>>, so unwrap is needed.
+    pub fn result(context: Option<&IFabricAsyncOperationContext>) -> crate::Result<T> {
+        let self_impl: &BridgeContext3<T> = unsafe { context.unwrap().as_impl() };
+        self_impl.consume_content()
+    }
+
+    // TODO: send and comsume is expected to happend accross threads.
+    // Even though we use a oneshot channel to send the signal,
```
jrahman, line 99: Oneshot channel should already have a memory barrier internally (you can confirm if you'd like), so this is ok as is

crates/libs/core/src/sync/cancel.rs
```diff
@@ -0,0 +1,729 @@
...
+                    Err(FabricErrorCode::OperationCanceled.into())
+                } else {
+                    panic!("content is consumed twice.")
+                }
+            }
+        }
+    }
+
+    fn set_complete(&self) {
+        self.is_completed.swap(&Cell::new(true));
```
jrahman, line 125: Any reason for swap() vs. .set()?

crates/libs/core/src/sync/cancel.rs
```diff
@@ -0,0 +1,729 @@
...
+    fn poll(mut self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Self::Output> {
+        // Poll the receiver first, if ready then return the output,
+        // else poll the cancellation token, if cancelled return the error,
+        // else return pending.
+        // There can be the case that cancellation wakes the waker, but receiver
+        // then got the result. The next poll will return received output rather
+        // than cancelled error.
+
+        // Try to receive the value from the sender
+        let innner =
```
jrahman, line 223: nit: innner -> inner

crates/libs/core/src/sync/cancel.rs
```diff
@@ -0,0 +1,729 @@
...
+                        }
+                        panic!("sender dropped without sending")
+                    }
+                }
+            }
+            Poll::Pending => {
+                // If the action is canceled we can safely stop and return canceled error.
+                if let Some(t) = &self.token {
+                    // this is cancel safe so we can poll it once and discard
+                    let fu = t.cancelled();
```
jrahman, line 248: I think you can flatten this by a couple levels of indentation:
```suggestion
        match (inner, self.token) {
            (Poll::Ready(Ok(data)), _) => {
                 Poll::Ready(Ok(data))
             },
             (Poll::Ready(Err(_)),  Some(t)) => {
                 if t.is_cancelled() {
                       // cancel the SF ctx
                      if let Err(e) = self.cancel_inner_ctx() {
                         return Poll::Ready(Err(e));
                      }
                     return Poll::Ready(Err(FabricErrorCode::OperationCanceled.into()));
                }
            }
            (Poll::Pending, Some(t) => {
                // If the action is canceled we can safely stop and return canceled error.
                // this is cancel safe so we can poll it once and discard
               let fu = t.cancelled();
```

crates/libs/core/src/sync/cancel.rs
```diff
@@ -0,0 +1,729 @@
...
+        if e.is_err() {
+            // In SF use case receiver should not be dropped by user.
+            // If it acctually dropped by user, it is ok to ignore because user
+            // does not want to want the value any more. But too bad SF has done
+            // the work to get the value.
+
+            // receiver should never be dropped if operation is not cancelled.
+            if let Some(t) = self.token {
+                if !t.is_cancelled() {
+                    panic!("receiver dropped.");
```
jrahman, line 292: panic!() or debug_assert!()? Panic!() feels a bit strong here, will the library/program malfunction if execution continues after this point?

crates/libs/core/src/sync/cancel.rs
```diff
@@ -0,0 +1,729 @@
...
+
+    // Set the SF ctx to hook up cancellation.
+    fn set_ctx(&mut self, ctx: IFabricAsyncOperationContext) {
+        let prev = self.ctx.replace(ctx);
+        assert!(prev.is_none());
+    }
+
+    // Cancels the inner SF operation if exists, and reset the ctx.
+    fn cancel_inner_ctx(&mut self) -> crate::Result<()> {
+        if let Some(ctx) = &self.ctx {
```
jrahman, line 193: Would we ever expect self.ctx to be None during normal/correct operation? ie. should it have a debug_assert!()?

rsajwani (approved):

crates/libs/core/src/sync/cancel.rs
youyuanwu: SF com layer always uses Option<&IFabricAsyncOperationCallback> as parameter. And this api's main use case is to be used in the SF com layer. I prefer to unwrap inside here rather than defer to the caller, and the number of unwraps in code will multiply by the number of SF apis. If later we decide to check the option instead of unwrap, the this api is significantly easier to change.

crates/libs/core/src/sync/cancel.rs
youyuanwu: This might be independent of oneshot. We set the result in the ctx in one thread, and then trigger the SF callback to notify the result is ready, and then the SF might use another thread to retrieve the result from ctx. 

The comment might have been copied from earlier BridgeContext implementation, which might have relied on oneshot.
The original comment might also indicate that if the Bridge is used in the Proxy layer in rust, then the Proxy layer has an oneshot channel  (which is in the SF callback), then we are covered. The only missing case are the SF callbacks which do not have any memory barriers inside.

I will keep the comment as is for now.

crates/libs/core/src/sync/cancel.rs
youyuanwu: Flattened. It has less indentations but more cases.

crates/libs/core/src/sync/cancel.rs
youyuanwu: Changed to debug_assert. Practically this should not happen.

crates/libs/core/src/sync/cancel.rs
youyuanwu: Added comments about this. This might happen when polling happened multiple times and we call cancelled inner ctx multiple times and the first time will empty the ctx after cancel.

youyuanwu:
In the current implementation, cancelling the token will preempt any rust future immediately that is wait for SF to complete. This might not be the desired behavior. The better behavior is that cancelling the token will not preempt futures immediately but still wait for SF to complete, and it is up to SF to respond to the cancellation request. If SF got the cancel signal, it finished the request fast with operation cancelled error code, the rust future should wait for the error code.

I will explore this in a future PR.

Issue #57, opened 2024-08-03
Support ServicePartitionResolver with cache


youyuanwu:
We need to implement a resolver on top of the fabric client similarliy in csharp: [ServicePartitionResolver Class (Microsoft.ServiceFabric.Services.Client) - Azure for .NET Developers | Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/api/microsoft.servicefabric.services.client.servicepartitionresolver?view=azure-dotnet)
 
Reference: [service-fabric-services-and-actors-dotnet/src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs at 6f7dc80b1f8e498df3bf2e1b0c41e75bc9ee9862 · microsoft/service-fabric-services-and-actors-dotnet (github.com)](https://github.com/microsoft/service-fabric-services-and-actors-dotnet/blob/6f7dc80b1f8e498df3bf2e1b0c41e75bc9ee9862/src/Microsoft.ServiceFabric.Services/Client/ServicePartitionResolver.cs#L627)

"
Service instances or replicas can move around from node to node at any time for multiple reasons. The service address resolved through ServicePartitionResolver may be stale by the time your client code attempts to connect. In that case again the client needs to re-resolve the address. Providing the previous
ResolvedServicePartition
indicates that the resolver needs to try again rather than simply retrieve a cached address."

The resolver needs to handle stale endpoint entry and register notification callbacks.

PR #60, opened 2024-08-13
[FabricClient] Support 'register service notification filter' api

youyuanwu:
Partial impl for #57 
FabricClient can register the service notification filter and unregister it.
See dotnet [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.fabric.fabricclient.servicemanagementclient.registerservicenotificationfilterasync?view=azure-dotnet)
The registration mechanism is used inside the dotnet implementation of [servicepartitionresolver](https://learn.microsoft.com/en-us/dotnet/api/microsoft.servicefabric.services.client.servicepartitionresolver?view=azure-dotnet)

The registered service to get notification will no longer need to pass the previous endpoint resolve result to force a refresh. This is tested in the stateful example app. But the notification seems to be not very fast (lag of a couple of seconds). But this notification utilizes FabricClient cache, and reduces the cluster api load.
I will compare with direct resolution next.

### DIFF
--- a/crates/libs/core/src/client/svc_mgmt_client.rs
+++ b/crates/libs/core/src/client/svc_mgmt_client.rs
@@ -12,11 +12,12 @@ use mssf_com::{
         FABRIC_PARTITION_KEY_TYPE_INVALID, FABRIC_PARTITION_KEY_TYPE_NONE,
         FABRIC_PARTITION_KEY_TYPE_STRING, FABRIC_REMOVE_REPLICA_DESCRIPTION,
         FABRIC_RESOLVED_SERVICE_ENDPOINT, FABRIC_RESTART_REPLICA_DESCRIPTION,
-        FABRIC_SERVICE_ENDPOINT_ROLE, FABRIC_SERVICE_PARTITION_KIND,
-        FABRIC_SERVICE_PARTITION_KIND_INT64_RANGE, FABRIC_SERVICE_PARTITION_KIND_INVALID,
-        FABRIC_SERVICE_PARTITION_KIND_NAMED, FABRIC_SERVICE_PARTITION_KIND_SINGLETON,
-        FABRIC_SERVICE_ROLE_INVALID, FABRIC_SERVICE_ROLE_STATEFUL_PRIMARY,
-        FABRIC_SERVICE_ROLE_STATEFUL_SECONDARY, FABRIC_SERVICE_ROLE_STATELESS, FABRIC_URI,
+        FABRIC_SERVICE_ENDPOINT_ROLE, FABRIC_SERVICE_NOTIFICATION_FILTER_DESCRIPTION,
+        FABRIC_SERVICE_PARTITION_KIND, FABRIC_SERVICE_PARTITION_KIND_INT64_RANGE,
+        FABRIC_SERVICE_PARTITION_KIND_INVALID, FABRIC_SERVICE_PARTITION_KIND_NAMED,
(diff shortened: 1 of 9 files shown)


crates/libs/core/src/client/svc_mgmt_client.rs
```diff
@@ -150,6 +189,47 @@ impl ServiceManagementClient {
...
     }
+
+    /// Remarks:
+    /// There is a cache of service endpoints in the client that gets updated by notifications
+    /// and this same cache is used to satisfy complaint based resolution requests
+    /// (see resolve_service_partition())). Applications that both register for notifications
+    /// and use complaint based resolution on the same client instance typically only need to
+    /// pass null for the ResolvedServicePartition argument during resolution.
+    /// This will always return the endpoints in the client cache updated by the latest notification.
+    /// The notification mechanism itself will keep the client cache updated when service endpoints change.
```
jrahman, line 200: Worth documenting typical/expected latency here

crates/libs/core/src/types/client/mod.rs
```diff
@@ -5,8 +5,50 @@
...
+    pub name: HSTRING,
+    pub flags: ServiceNotificationFilterFlags,
+}
+
+impl From<&ServiceNotificationFilterDescription>
+    for FABRIC_SERVICE_NOTIFICATION_FILTER_DESCRIPTION
+{
+    fn from(value: &ServiceNotificationFilterDescription) -> Self {
+        Self {
+            Name: FABRIC_URI(value.name.as_ptr() as *mut u16),
```
jrahman, line 49: Does FABRIC_URI internally perform a deep copy on value pointed to by the pointer passed as a parameter?

jrahman (approved):

crates/samples/echomain/src/test.rs
```diff
@@ -67,14 +135,63 @@ async fn test_fabric_client() {
...
-        partition_id: single.id,
-        replica_or_instance_id: stateless_replica.instance_id,
+    // register service notification filter
+    let filter_handle = {
+        let desc = ServiceNotificationFilterDescription {
+            name: service_uri.clone(),
+            flags: ServiceNotificationFilterFlags::NamePrefix,
+        };
+        // register takes more than 1 sec.
+        mgmt.register_service_notification_filter(&desc, Duration::from_secs(10))
```
rsajwani, line 146: So basically , the caller need to retry if they don't get expected count ?? Notification will eventually catchup on that .. Any possibility that notification can be lost .. Is it best effort?

rsajwani (approved):

crates/samples/echomain/src/test.rs
youyuanwu: Caller needs to retry if the addr is not expected, including the number of addrs are not correct.
AFAIK, new addrs should eventually show up. I will keep looking if there are better ways, since I observed some slowness.

crates/libs/core/src/client/svc_mgmt_client.rs
youyuanwu: I will send another PR to compare this solution with brute force retry and added the results then.

crates/libs/core/src/types/client/mod.rs
youyuanwu: It is not the deep copy. The SF raw type returned from this conversion should have the same lifetime as the original value. I will add a note in the next PR.

PR #61, opened 2024-08-13
[FabricClient] Service notification perf comparison

youyuanwu:
Added a test to compare service notification performance with the brute force complain based resolution.
In the test notification is observed to be 1 to 4 seconds slower than the brute force way, and the typical lag for brute force is 6 seconds. So the notification takes about 7 to 10 seconds.


jrahman (approved):

rsajwani (commented):
LGTM

PR #62, opened 2024-08-15
[sync] Cancellation waits for SF task to complete

youyuanwu:
Currently when cancelling the SF operation proxy operation, the rust future is short circuited when the Cancellation Token is cancelled. This means that rust code does not wait for SF operation to complete or react to the cancel signal.
This PR changes the behavior to let SF operation proxy wait for sf operation to complete. If the operation is cancelled, wait for sf operation's response. It is sf operation's responsibility to react to the signal, and complete faster (with and error code).
If user wants to discard SF task result and short circuit without waiting, user needs to use select! of the sf operation and the cancellation token (not recommended or yet tested).

### DIFF
--- a/crates/libs/core/src/sync/cancel.rs
+++ b/crates/libs/core/src/sync/cancel.rs
@@ -217,8 +217,11 @@ impl<T> Future for FabricReceiver2<T> {
     type Output = crate::Result<T>;
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         // Poll the receiver first, if ready then return the output,
-        // else poll the cancellation token, if cancelled return the error,
-        // else return pending.
+        // else poll the cancellation token, if cancelled propergate the cancel to SF ctx,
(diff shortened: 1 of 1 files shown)


rsajwani:
can you point me out where is this part in the code 

>>It is sf operation's responsibility to react to the signal, and complete faster (with and error code).

crates/libs/core/src/sync/cancel.rs
```diff
@@ -466,13 +479,14 @@ mod test {
...
         ) -> crate::Result<String> {
             if delay.is_zero() {
                 return Ok(self.get_data());
             }
-            match token {
-                Some(t) => {
+            match (token, ignore_cancel) {
+                (Some(t), false) => {
                     select! {
                         _ = t.cancelled() => {
```
youyuanwu, line 491: @rsajwani For example there in the implementation, the cancel signal is received and we cancel the operation immediately without processing the request with an error code.
If we ignore this cancel signal, as we do in some other test branch, the delay would happen irrespective to the signal.

rsajwani (approved):

PR #65, opened 2024-08-23
Cancellation: Add CancellationToken to all APIs

youyuanwu:
Cancellation is supported since #58 .
This PR adds CancellationToken to all apis.


rsajwani:
are you going to delete proxy in later. or will we keep both proxy and proxy2

rsajwani (approved):

youyuanwu:
> are you going to delete proxy in later. or will we keep both proxy and proxy2

I deleted the original proxy wrapper. Will do the rename later.

PR #67, opened 2024-08-29
Fix tokio_async disabled build; Support bringing your own import libraries

cgettys-microsoft:
* Fix tokio_async disabled build by conditionally compiling out some more async related code
* Add new sample library to ensure we're able to compile with default features disabled. Not doing the full 2^3 combinations, but right now, the 3 are completely orthgonal as far as I know, so this is a good start
* add new default-on bundled_lib option; if disabled, consumer will need to provide import libs in their own build script or similar. This allows using custom import libs. It's not pretty, but should work. For now, only address windows bring own scenario.

### DIFF
--- a/crates/libs/com/src/lib.rs
+++ b/crates/libs/com/src/lib.rs
@@ -11,9 +11,9 @@ pub mod ServiceFabric;
 // expose mod directly
 pub use ServiceFabric::*;
 
-#[cfg(all(target_os = "windows", feature="bundled_import_libs"))]
+#[cfg(all(target_os = "windows", feature = "bundled_import_libs"))]
 pub use mssf_metadata;
 
 // In linux force to pull in pal lib for linking
-#[cfg(all(target_os = "linux", feature="bundled_import_libs"))]
+#[cfg(all(target_os = "linux", feature = "bundled_import_libs"))]
 extern crate mssf_pal;
--- a/crates/libs/com/src/lib.rs
+++ b/crates/libs/com/src/lib.rs
@@ -11,9 +11,9 @@ pub mod ServiceFabric;
 // expose mod directly
 pub use ServiceFabric::*;
 
-#[cfg(all(target_os = "windows", feature = "bundled_import_libs"))]
+#[cfg(all(target_os = "windows", feature = "bundled_libs"))]
 pub use mssf_metadata;
 
 // In linux force to pull in pal lib for linking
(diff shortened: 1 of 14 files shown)


crates/libs/com/src/lib.rs
```diff
@@ -11,9 +11,9 @@ pub mod ServiceFabric;
 // expose mod directly
 pub use ServiceFabric::*;
 
-#[cfg(target_os = "windows")]
+#[cfg(all(target_os = "windows", feature = "bundled_import_libs"))]
 pub use mssf_metadata;
 
 // In linux force to pull in pal lib for linking
-#[cfg(target_os = "linux")]
+#[cfg(all(target_os = "linux", feature = "bundled_import_libs"))]
```
youyuanwu, line 18: Why pal needs to have bundled_import_libs feature?
linux does not have import libs.

crates/samples/only-bundled-libs/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# copy files to build folder to form a code package.
+
+add_custom_target(build_rust_sample_only-bundled-libs ALL
+    COMMAND ${cargo_exe} build -p only_bundled_libs
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+    DEPENDS build_fabric_rust_pal
+)
+
+set(_pkg_root ${CMAKE_BINARY_DIR}/only-bundled-libs_root)
```
youyuanwu, line 9: unused variables can be removed.

crates/samples/only-bundled-libs/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+//! The known use case for no-default-features is as part of existing apps that already have lifecycle management
+//! in another language
+//! So building a sample app is somewhat more involved
+//! For now, we just check that we can compile mssf-core and mssf-com without any features except for the bundled import libs.
```
youyuanwu, line 4: What is testing? The motivation is unclear.

cgettys-microsoft: Improved the comment

crates/samples/only-bundled-libs/CMakeLists.txt
cgettys-microsoft: Sorry, rusty (ha!) on CMake - should have realized those weren't necessary ones since they didn't start with CMAKE_

youyuanwu (approved):

PR #80, opened 2024-10-17
[FabricRuntime] impl get code package info from activation context

youyuanwu:
code package info like work dir, log dir etc can be retrieved from activation context.
activation context is refactored to separate mod.

### DIFF
--- a/crates/libs/core/src/runtime/activation_context.rs
+++ b/crates/libs/core/src/runtime/activation_context.rs
@@ -1,0 +1,99 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+use mssf_com::FabricRuntime::IFabricCodePackageActivationContext6;
+
+use crate::{strings::HSTRINGWrap, types::EndpointResourceDesc, Error, HSTRING, PCWSTR};
+
+use super::config::ConfigurationPackage;
+
+pub struct CodePackageActivationContext {
+    com_impl: IFabricCodePackageActivationContext6,
+}
+
+/// Info from the context
+#[derive(Debug, Clone)]
+pub struct CodePackageInfo {
+    pub context_id: HSTRING,
+    pub code_package_name: HSTRING,
+    pub code_package_version: HSTRING,
+    pub work_directory: HSTRING,
+    pub log_directory: HSTRING,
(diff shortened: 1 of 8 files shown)


crates/libs/core/src/types/runtime/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+use crate::HSTRING;
+
+#[derive(Debug)]
+pub struct EndpointResourceDesc {
+    pub name: HSTRING,
+    pub protocol: HSTRING,
+    pub r#type: HSTRING,
+    pub port: u32,
+    pub certificate_name: HSTRING,
+    //pub Reserved: *mut ::core::ffi::c_void,
```
rsajwani, line 19: future use??

rsajwani (approved):

crates/libs/core/src/runtime/activation_context.rs
```diff
@@ -0,0 +1,99 @@
...
+
+use crate::{strings::HSTRINGWrap, types::EndpointResourceDesc, Error, HSTRING, PCWSTR};
+
+use super::config::ConfigurationPackage;
+
+pub struct CodePackageActivationContext {
+    com_impl: IFabricCodePackageActivationContext6,
+}
+
+/// Info from the context
```
jrahman, line 16:
```
/// Code package information is propagated here from the ActivationContext
/// provided when the service is started. This information can be used to setup
/// log directories and other resources needed which require information
/// about the code package and the local runtime environment.
```

crates/samples/echomain/src/main.rs
```diff
@@ -39,11 +39,14 @@ fn main() -> mssf_core::Result<()> {
...
         wait_for_debugger();
     }
-    let actctx = ActivationContext::create().inspect_err(|e| {
+    let actctx = CodePackageActivationContext::create().inspect_err(|e| {
         error!("Fail to create activation context: {e}");
     })?;
     validate_configs(&actctx);
 
+    let code_info = actctx.get_code_package_info();
+    info!("code package info: {:?}", code_info);
```
jrahman, line 48: Debugging?

jrahman (approved):

crates/libs/core/src/types/runtime/mod.rs
youyuanwu: Removed.

crates/samples/echomain/src/main.rs
youyuanwu: Example app output info is ok.

PR #85, opened 2024-10-26
Add new methods to extract fields in PartitionLoadInformation

OscarTHZhang:


crates/libs/core/src/types/client/partition.rs
```diff
@@ -233,4 +234,25 @@ impl PartitionLoadInformation {
...
+    /// Returns the partition id
+    pub fn partition_id(&self) -> GUID {
+        self.partition_id
+    }
+
+    /// Returns a list of load metric reports for primary replica
+    pub fn primary_load_metric_reports(&self) -> Vec<LoadMetricReport> {
+        self
+            .primary_load_metric_reports
+            .iter()
```
youyuanwu, line 247: Just return the PrimaryLoadMetricReportList. make the field private.
mssf api prefers not use Vec or other types that requires dynamic alloc for list.

youyuanwu (changes requested):
Fix API and CI.

OscarTHZhang:
In that case, probably for simplicity we can stick with the current implementation. Just treating the struct as a C-style struct with public members

PR #91, opened 2024-11-11
Replicator documentation

youyuanwu:
Add unofficial documentation for replicator apis.

### DIFF
--- a/crates/libs/core/src/runtime/stateful.rs
+++ b/crates/libs/core/src/runtime/stateful.rs
@@ -35,6 +35,10 @@ pub trait StatefulServiceFactory {
 pub trait LocalStatefulServiceReplica: Send + Sync + 'static {
     /// Opens an initialized service replica so that additional actions can be taken.
     /// Returns PrimaryReplicator that is used by the stateful service.
+    /// Note:
+    /// Most user calls IFabricStatefulServicePartition.CreateReplicator instead of
+    /// writing their own replicator (TODO: not supported in mssf yet),
+    /// or use FabricCreateKeyValueStoreReplica.
     async fn open(
         &self,
         openmode: OpenMode,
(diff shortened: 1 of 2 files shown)


crates/libs/core/src/runtime/stateful.rs
```diff
@@ -95,8 +103,17 @@ impl From<&IFabricStatefulServicePartition> for StatefulServicePartition {
...
+    /// in ReplicaInformation.
+    /// Remarks:
+    /// Replicator does not have an assigned role yet and should setup listening endpoint.
+    async fn open(&self, cancellation_token: CancellationToken) -> crate::Result<crate::HSTRING>;
     async fn close(&self, cancellation_token: CancellationToken) -> crate::Result<()>;
+
+    /// Change the replicator role.
+    ///
+    /// Remarks:
+    /// Replicator change_role is called before Replica change_role.
```
jrahman, line 116: Is there any broader commentary to offer on the relationship between `Replica` and `Replicator`?

crates/libs/core/src/runtime/stateful.rs
```diff
@@ -111,43 +128,132 @@ pub trait LocalReplicator: Send + Sync + 'static {
...
+    // along with the current write quorum. In addition there can be a previous configuration
+    // which represents the set of replicas that were in the previous configuration.
+    // When there is both a current and previous configuration the replicator must ensure that
+    // writes are acknowledeged by a write quroum of both configurations.
+
+    /// Informs the replicator there there is a current configuration and a previous configuration.
+    /// Called on primary to inform the set of active secondary replicas that may
+    /// begin to catchup. Idle secondary replicas are not included here.
+    ///
+    /// The total number of replica marked with must_catchup will not exceed the write quorum.
```
jrahman, line 190: What is the significance of the `must_catchup` flag when provided as `true`?

crates/libs/core/src/runtime/stateful.rs
```diff
@@ -111,43 +128,132 @@ pub trait LocalReplicator: Send + Sync + 'static {
...
+    /// Informs the replicator about the current replica set configuration, and there
+    /// is no longer a previous configuration.
+    /// Remarks:
+    /// Replicas here are not marked as must_catchup.
+    fn update_current_replica_set_configuration(
+        &self,
+        currentconfiguration: &ReplicaSetConfig,
+    ) -> crate::Result<()>;
+
+    /// Called on primary to wait for the quorum of replicas to catch up, before
```
jrahman, line 214: Might be worth defining what "catch up" means from the SF PoV

crates/libs/core/src/runtime/stateful.rs
youyuanwu: If the catchup mode is write, it is upto the replicator to pick a quorum of the replicas that includes the must_catch up replica. Wait_for_catchup_quorum complete can return if the quorum of replica has caught up, and can leave other slow replica still running in background.

crates/libs/core/src/runtime/stateful.rs
youyuanwu: Added in the remarks.

crates/libs/core/src/runtime/stateful.rs
youyuanwu: Replica is for setting up listener for user connections and manage its life cycle. 
Replicator is for replication listener and LSN management in the background.

rsajwani (approved):

Issue #92, opened 2024-11-12
Support all actions for IFabricStatefulServicePartition


youyuanwu:
Only a subset of functions for IFabricStatefulServicePartition is exposed in mssf so far.

PR #95, opened 2024-11-12
Impl and expose IFabricStatefulServicePartition3 apis

youyuanwu:
Expose IFabricStatefulServicePartition3 apis in Rust.
Some complicated apis are left out in this PR and will be added in the future.
#92

### DIFF
--- a/crates/libs/core/src/runtime/stateful_proxy.rs
+++ b/crates/libs/core/src/runtime/stateful_proxy.rs
@@ -10,20 +10,22 @@ use std::ffi::c_void;
 
 use mssf_com::FabricRuntime::{
     IFabricPrimaryReplicator, IFabricReplicator, IFabricReplicatorCatchupSpecificQuorum,
-    IFabricStatefulServiceReplica,
+    IFabricStatefulServicePartition3, IFabricStatefulServiceReplica,
 };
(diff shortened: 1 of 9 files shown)


crates/libs/core/src/runtime/stateful_proxy.rs
```diff
@@ -306,3 +308,95 @@ impl PrimaryReplicator for PrimaryReplicatorProxy {
...
+        unsafe { self.com_impl.ReportMoveCost(move_cost.into()) }
+    }
+
+    /// Remarks:
+    /// The health information describes the report details, like the source ID, the property,
+    /// the health state and other relevant details. The partition uses an internal health client
+    /// to send the reports to the health store. The client optimizes messages to Health Manager
+    /// by batching reports per a configured duration (Default: 30 seconds). If the report has high priority,
+    /// you can specify send options to send it immediately.
+
```
OscarTHZhang, line 382: nit: extra white space

youyuanwu: This is intentional the remark applies to the following 2 apis but not only to the first one.

OscarTHZhang (approved):

PR #96, opened 2024-11-15
[FabricRuntime] Support Package change notification callbacks

cgettys-microsoft:
End user applications probably want to handle package change notifications.

This PR currently tackles only ConfigurationPackage.

If you're happy with the approach, I'll do the same exact implementation (edit: in additional PRs) for CodePackage and DataPackage and that will close out #97.


crates/libs/core/src/client/package_change/config.rs
```diff
@@ -0,0 +1,146 @@
...
+    IFabricConfigurationPackageChangeHandler, IFabricConfigurationPackageChangeHandler_Impl,
+};
+
+use crate::runtime::config::ConfigurationPackage;
+
+use super::PackageChangeType;
+
+#[derive(Debug, Clone)]
+pub struct ConfigurationPackageChangeEvent {
+    pub change_type: PackageChangeType,
```
youyuanwu, line 17: Maybe change_type could be an enum.

crates/libs/core/src/client/mod.rs
```diff
@@ -109,6 +119,7 @@ impl FabricClientBuilder {
         Self {
             sn_handler: None,
             cc_handler: None,
+            conf_pkg_handler: None,
```
youyuanwu, line 122: Where is this passed to SF? Seems like it is dropped after builder is dropped.

crates/libs/core/src/client/package_change/config.rs
cgettys-microsoft: It is, in crates/libs/core/src/client/package_change/mod.rs
```
pub enum PackageChangeType {
    Addition,
    Removal,
    Modification,
}
```
Do you mean I could make ConfigurationPackageChangeEvent an enum with actual fields?? Yeah, I could
```
pub enum PackageChangeEvent<T> {
    Addition { next_package: T },
    Removal { next_package: T }
    Modification { previous_package: T, next_package: T},
}
```
That's what I do for diffs inside my library

youyuanwu: Yes. Enum with fields.

crates/libs/core/src/client/mod.rs
cgettys-microsoft: It wasn't. Added a wrapper to handle it.

cgettys-microsoft: But said wrapper isn't complete or functional. Trying to wrap my head around the interaction of Rust lifetime rules and COM....

crates/libs/core/src/runtime/package_change/config.rs
```diff
@@ -0,0 +1,147 @@
...
+
+/// This struct ensures that the handle is retained and deregistered before the implementation is dropped
+struct ConfigurationPackageChangeEventHandlerManager
+{
+    activation_ctx: IFabricCodePackageActivationContext6,
+    implementation: IFabricConfigurationPackageChangeHandler,
+    handle: ConfigurationPackageChangeHandlerId
+}
+
+impl ConfigurationPackageChangeEventHandlerManager
```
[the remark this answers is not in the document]
youyuanwu: And for simplicity, just return a handle and let caller unregister it: https://github.com/Azure/service-fabric-rs/blob/2c8933b52016bb9f864f1e7aa55a49df2cb27d85/crates/libs/core/src/client/svc_mgmt_client.rs#L226C1-L232C1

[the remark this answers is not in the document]
cgettys-microsoft: Provided a simpler struct for caller to drop. Would rather not risk leaks if we can help it. I see why you did what you did, and lifetime bound would also work, but I think this is cleaner from a usage perspective - drop handle, everything is cleaned up.

crates/libs/core/src/runtime/package_change/config.rs
```diff
@@ -0,0 +1,143 @@
...
+{
+    fn on_change(&self, change: &ConfigurationPackageChangeEvent) -> crate::Result<()> {
+        (self.f)(change)
+    }
+}
+
+/// This struct ensures that the handle is retained and deregistered before the implementation is dropped
+pub struct ConfigurationPackageChangeCallbackHandle {
+    activation_context: IFabricCodePackageActivationContext6,
+    handle: i64,
```
youyuanwu, line 116: Just expose actctx function to return and handle and user needs to unregister. Not deviate too much from the COM api. I think the runtime will auto cleanup the handle at drop time.

youyuanwu: U can add another type for auto drop. On top of the handle.

cgettys-microsoft: Better?

cgettys-microsoft: Separated the behavior.

cgettys-microsoft: let me know which you'd like

crates/libs/core/src/runtime/package_change/config.rs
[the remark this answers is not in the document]
youyuanwu: There is already another handle in fabric client with similar intention: https://github.com/Azure/service-fabric-rs/blob/0f2e5e9f734ccb8a3ebd47303182e4f63e074152/crates/libs/core/src/client/svc_mgmt_client.rs#L232

There should be 2 separate concerns in mssf. 1) Exposing the api in rust and make it clear to SF user what this api corresponds to in COM. 2) make the rust api safe.
The code currently achieves 2), but not 1).

crates/samples/no_default_features/src/lib.rs
```diff
@@ -10,11 +10,26 @@
...
 fn test_fn() {
-    // Make sure we link something
-    //
-    let my_ctx = CodePackageActivationContext::create();
-    my_ctx.unwrap();
+    let my_ctx = CodePackageActivationContext::create().unwrap();
+
+    // One might wish to use such a callback to e.g. trigger custom handling of configuration changes
+    // This doesn't require the config feature to be enabled
+    let _handler = my_ctx.register_config_package_change_handler( |c|
```
youyuanwu, line 21: U should move this to one of the example apps.

cgettys-microsoft: Huh? this is one of the samples

crates/libs/core/src/runtime/package_change/config.rs
cgettys-microsoft: Removed auto-drop for now, can always add it later if we want it.

crates/libs/core/src/runtime/package_change/config.rs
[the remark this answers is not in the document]
cgettys-microsoft: Did it your way

crates/samples/no_default_features/src/lib.rs
cgettys-microsoft: Oh, you mean one of the full applications.
Yeah, that could make sense.
I do want to ensure this continues to build without those features though.

cgettys-microsoft: I think the logical example to add it to would be echomain

As we could demonstrate how to hook it up to trigger a Config reload

cgettys-microsoft: Basically this but with our source:
https://github.com/rust-cli/config-rs/blob/main/examples/watch/main.rs

crates/libs/core/src/runtime/package_change/config.rs
```diff
@@ -0,0 +1,168 @@
...
+    T: ConfigurationPackageChangeEventHandler,
+{
+    fn OnPackageAdded(
+        &self,
+        _source: Option<&mssf_com::FabricRuntime::IFabricCodePackageActivationContext>,
+        configpackage: Option<&mssf_com::FabricRuntime::IFabricConfigurationPackage>,
+    ) {
+        let new_package = ConfigurationPackage::from_com(configpackage.unwrap().clone());
+        let event = ConfigurationPackageChangeEvent::Addition { new_package };
+        // TODO: unwrap, or should we change the return type of the lambda to be the empty type?
```
cgettys-microsoft, line 52: Thoughts?

crates/samples/no_default_features/src/lib.rs
cgettys-microsoft: Given I'd like to have this build here as well, can I defer that to a second PR?

youyuanwu: U can keep it here or add to the working app.
But note that this code here cannot run because it is not launched by SF, and we are not packaging it in an app.

crates/libs/core/src/runtime/package_change/config.rs
youyuanwu: Keep aligned with the COM api. SF does not handle failure return, so Rust api should match it. i.e. no return type

youyuanwu (approved):

crates/libs/core/src/runtime/package_change/config.rs
cgettys-microsoft: Fixed

crates/samples/no_default_features/src/lib.rs
cgettys-microsoft: Added it to echomain as well (but haven't had a chance to test that app yet)
Right, point of this is to make sure we can package w/o tokio

PR #101, opened 2024-11-22
Support reporting app health

youyuanwu:
App health can be reported from the activation context.
It shows in the SF health store.
image

### DIFF
--- a/crates/libs/core/src/types/runtime/health.rs
+++ b/crates/libs/core/src/types/runtime/health.rs
@@ -1,0 +1,72 @@
+use mssf_com::FabricTypes::{FABRIC_HEALTH_INFORMATION, FABRIC_HEALTH_REPORT_SEND_OPTIONS};
+use windows_core::PCWSTR;
+
+use crate::{strings::HSTRINGWrap, types::HealthState, HSTRING};
+
+pub type SequenceNumber = i64;
+
+/// FABRIC_HEALTH_INFORMATION
+#[derive(Debug, Clone)]
+pub struct HealthInformation {
+    pub source_id: HSTRING,
+    pub property: HSTRING,
+    pub time_to_live_seconds: u32,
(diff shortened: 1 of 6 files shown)


crates/libs/core/src/types/runtime/health.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// FABRIC_HEALTH_INFORMATION
+#[derive(Debug, Clone)]
+pub struct HealthInformation {
+    pub source_id: HSTRING,
+    pub property: HSTRING,
+    pub time_to_live_seconds: u32,
+    pub state: HealthState,
+    pub description: HSTRING,
+    pub sequence_number: SequenceNumber,
```
jrahman, line 16: What is the significance of the sequence number here? Assuming the application populates it to ensure ordering in the event or retries or misordered messages on the network?

crates/libs/core/src/types/runtime/health.rs
```diff
@@ -0,0 +1,72 @@
...
+
+use crate::{strings::HSTRINGWrap, types::HealthState, HSTRING};
+
+pub type SequenceNumber = i64;
+
+/// FABRIC_HEALTH_INFORMATION
+#[derive(Debug, Clone)]
+pub struct HealthInformation {
+    pub source_id: HSTRING,
+    pub property: HSTRING,
```
jrahman, line 12: Semantically, what sort of information should be placed in the source_id and property fields? The HSTRING type itself and even the names are not super clear what the expectations are for these values.

jrahman (approved):

rsajwani (approved):

crates/libs/core/src/types/runtime/health.rs
youyuanwu: added documentation

PR #107, opened 2024-12-05
Upgrade windows-rs to 0.58 version

youyuanwu:
This is a significant improvement for trimming many dependencies.
* On linux, windows system crate is not used anymore.
* Pal c function implementation is removed. windows crate on linux does not reference those c functions anymore.
* libc dependency removed. All alloc should use rust std now.
* windows-core does not support strings anymore on linux, so we provide a minimal implementation of string ptr types and HSTRING (to be renamed) for code to compile on linux. 
* mssf-pal crate is turned into a trimmed version of windows_core that works on linux and windows.
This completes the long waited version update previously blocked by the windows_core string support drop. This prepares better for the next windows-bindgen rewrite which will come soon.


youyuanwu:
There is an ~14 seconds build time reduction for linux for this repo.

OscarTHZhang (approved):

PR #109, opened 2024-12-10
Dynamically load SF apis without linking

youyuanwu:


youyuanwu:
Might be able to do this using https://doc.rust-lang.org/nightly/std/sync/struct.LazyLock.html.
Will do it in a separate PR.

rsajwani (approved):

PR #112, opened 2024-12-17
Remove lld linker option and dependencies

youyuanwu:
Originally lld was used for linking with SF shared libs due to SF shared libs incompatibility with ld.
Now SF shared libs are dynamically loaded instead of linked, there is no need to use lld anymore.
Removes cargo lld linker option.


cgettys-microsoft:
Looks good to me, but I don't have reviewer powers in your repo, so I can't provide an approval obviously :)

OscarTHZhang (approved):

PR #121, opened 2025-01-09
Add a AutoConfigurationPackageChangeCallbackHandle RAII Type

cgettys-microsoft:
Add a useful RAII type to ensure handle is unregistered when type is dropped.

With more work, it could handle lambdas that aren't 'static too, but leaving that for future.

### DIFF
--- a/crates/libs/core/src/runtime/package_change/config.rs
+++ b/crates/libs/core/src/runtime/package_change/config.rs
@@ -8,7 +8,7 @@ use mssf_com::FabricRuntime::{
     IFabricConfigurationPackageChangeHandler, IFabricConfigurationPackageChangeHandler_Impl,
 };
 
-use crate::runtime::config::ConfigurationPackage;
+use crate::runtime::{config::ConfigurationPackage, CodePackageActivationContext};
 
 use super::ConfigurationPackageChangeEvent;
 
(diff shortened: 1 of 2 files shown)


crates/libs/core/src/runtime/package_change/config.rs
```diff
@@ -119,3 +121,38 @@ impl ConfigurationPackageChangeCallbackHandle {
...
+    /// Service Fabric Activation Context
+    activation_ctx: CodePackageActivationContext,
+    /// Handle to deregister on drop
+    handle: Option<ConfigurationPackageChangeCallbackHandle>,
+}
+
+impl AutoConfigurationPackageChangeCallbackHandle {
+    /// Register a new handle for the provided lambda.
+    /// Clones (e.g. adjusts reference count) on activation_ctx
+    pub fn new<T>(activation_ctx: &CodePackageActivationContext, handler: T) -> crate::Result<Self>
```
youyuanwu, line 138: U can pass the ConfigurationPackageChangeCallbackHandle instead of the function.
Or u can add the into_auto() function to ConfigurationPackageChangeCallbackHandle to convert directly to the auto type.

youyuanwu (approved):

crates/libs/core/src/runtime/package_change/config.rs
cgettys-microsoft: Sure, or better than into_auto, I could just implement From. WIll revisit this tomorrow

PR #122, opened 2025-01-09
Allow using FabricClient in non-tokio projects and refactor interface.

cgettys-microsoft:
While you still end up needing to drop back down to the com level at least today, not having to write casts by hand is nice.

Also, it's unnecessary to clone the COM level interface all the time. Make FabricClient own the high level wrappers; if you need it to live longer, you can clone it. This is a breaking change, so may warrant a version bump.

### DIFF
--- a/crates/libs/core/src/client/query_client.rs
+++ b/crates/libs/core/src/client/query_client.rs
@@ -41,8 +41,8 @@ pub struct QueryClient {
 // Internal implementation block
 // Internal functions focuses on changing SF callback to async future,
 // while the public apis impl focuses on type conversion.
+#[cfg(feature = "tokio_async")]
 impl QueryClient {
-    #[cfg(feature = "tokio_async")]
     pub fn get_node_list_internal(
         &self,
         query_description: &FABRIC_NODE_QUERY_DESCRIPTION,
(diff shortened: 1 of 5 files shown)


crates/libs/core/src/client/query_client.rs
```diff
@@ -38,6 +42,7 @@ pub struct QueryClient {
 // Internal functions focuses on changing SF callback to async future,
 // while the public apis impl focuses on type conversion.
 impl QueryClient {
```
youyuanwu, line 44: Can u cfg feature on the impl block?

cgettys-microsoft: I can, but then we end up with multiple impl blocks in some places

crates/libs/core/src/client/query_client.rs
```diff
@@ -109,10 +117,15 @@ impl QueryClient {
 }
 
 impl QueryClient {
+    pub fn get_com(&self) -> IFabricQueryClient10 {
```
cgettys-microsoft, line 120: e.g. woul dneed to split this impl block up

crates/libs/core/src/client/query_client.rs
cgettys-microsoft: Done

crates/libs/core/src/client/query_client.rs
cgettys-microsoft: Done

youyuanwu (approved):

PR #125, opened 2025-01-11
Make the code compile on macOS

OscarTHZhang:
With the optimization that build does not require Service Fabric runtime installed, building on any major platform is allowed. Here adding a small fix to make the code compile on macOS, though Service Fabric does not run natively on macOS.


crates/libs/core/src/debug/mod.rs
```diff
@@ -18,3 +18,7 @@ pub fn wait_for_debugger() {
 
 #[cfg(target_os = "linux")]
```
youyuanwu, line 19: cfg(not(target_os = "windows")) would work as well.

youyuanwu (approved):

PR #126, opened 2025-01-13
Replace from_com and into_com with From implementations

cgettys-microsoft:
Given that these from_com and into_com methods are all safe anyway, there's no good reason not to use From instead

Rewrite them as ```impl From<ITYPE> for TYPE``` and ```impl From<ITYPE> for TYPE```.
In a few cases, the second (which would have been into_com before) did not exist, but made sense, so I added them.
As per general Rust guidance, implementing From, as that also results in the blanket ```Into``` implementation.

In the vast majority of cases, ownership was required, so I went with ```impl From<ITYPE> for TYPE``` instead of ```impl From<&ITYPE> for TYPE```, as in some cases, this will save a reference count increment, and in the other case, it just moves where .clone is called, and those implementations won't be called frequently outside the library.

This also saved a few unnecessary clones inside the library that we appear to have had by accident.

Note in a few cases this makes it possible to construct the individual client wrappers outside the library (where we didn't have pub on from_com before). This is in my view a feature.

This will require any code outside this library to replace from_com with from and into_com with .into, but that shouldn't be a big deal.

I left ```FabricClient::from_com``` alone because the comment says the method is intended to be private eventually.

### DIFF
--- a/crates/libs/core/src/client/connection.rs
+++ b/crates/libs/core/src/client/connection.rs
@@ -27,8 +27,8 @@ pub struct GatewayInformationResult {
     pub node_name: crate::WString,
 }
 
-impl GatewayInformationResult {
-    fn from_com(com: &IFabricGatewayInformationResult) -> Self {
+impl From<&IFabricGatewayInformationResult> for GatewayInformationResult {
+    fn from(com: &IFabricGatewayInformationResult) -> Self {
         let info = unsafe { com.get_GatewayInformation().as_ref().unwrap() };
         Self {
             node_address: WStringWrap::from(info.NodeAddress).into(),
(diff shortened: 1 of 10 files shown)


rust-toolchain.toml
```diff
@@ -2,5 +2,5 @@
 # cargo + rustup will use this to consistently build the project
 # with the same version across all checkouts and environments
 [toolchain]
-channel = "1.84.0"
+channel = "ms-1.84"
```
OscarTHZhang, line 5: msrustup is Microsoft internal toolchain. Shouldn't be reflect on a public open-source repo

cgettys-microsoft: Whoops, didn't mean to include that in this change, let me fix.

cgettys-microsoft: Fixed

youyuanwu (approved):

PR #138, opened 2025-03-06
draft: Add support for configuring Fabric Client Settings and non-default auth settings

cgettys-microsoft:
* I have a use case for non-default authentication
* And may also want to configure some client settings
The COM APIs are quite unfriendly. Add idiomatic high-level interfaces

### DIFF
--- a/crates/libs/core/src/types/common/security_credentials/windows_credentials.rs
+++ b/crates/libs/core/src/types/common/security_credentials/windows_credentials.rs
@@ -5,9 +5,15 @@
 #![deny(unsafe_op_in_unsafe_fn)]
 #![deny(clippy::undocumented_unsafe_blocks)]
 
-use std::{ffi::c_void, ptr::{self, addr_of_mut}};
+use std::{
+    ffi::c_void,
+    ptr::{self, addr_of_mut},
+};
 
-use mssf_com::FabricTypes::{FABRIC_SECURITY_CREDENTIALS, FABRIC_SECURITY_CREDENTIAL_KIND_WINDOWS, FABRIC_WINDOWS_CREDENTIALS};
+use mssf_com::FabricTypes::{
+    FABRIC_SECURITY_CREDENTIALS, FABRIC_SECURITY_CREDENTIAL_KIND_WINDOWS,
+    FABRIC_WINDOWS_CREDENTIALS,
+};
 use windows_core::{WString, PCWSTR};
 
 use super::{FabricProtectionLevel, FabricSecurityCredentialKind};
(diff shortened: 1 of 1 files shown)


crates/libs/core/src/client/mod.rs
```diff
@@ -80,7 +82,19 @@ fn create_local_client_internal<T: Interface>(
         }
     }
     // if params are right, client should be created. There is no network call involved during obj creation.
-    .expect("failed to create fabric client")
+    .expect("failed to create fabric client");
+
+    // TODO: all interfaces this could be used for implement IFabricClientSettings2, right?
```
cgettys-microsoft, line 87: TODO

crates/libs/core/src/client/mod.rs
```diff
@@ -80,7 +82,19 @@ fn create_local_client_internal<T: Interface>(
...
+    if let Some(desired_settings) = client_settings {
+        let setting_interface = client
+            .clone()
+            .cast::<IFabricClientSettings2>()
+            .expect("failed to cast fabric client to IFabricClientSettings2");
+        // SAFETY: setting_interface implements the required COM interface.
+        let existing_settings = FabricClientSettings::get_from_com(&setting_interface);
```
cgettys-microsoft, line 94: There are three approaches I could take:
A) Retrieve the current settings from SF, and merge in any user provided settings
B) Hard-code defaults for all settings (ugly, brittle). and merge in any user provided settings on top
C) Require the user to specify all settings.
I'm planning on going with Option A

crates/libs/core/src/types/client/security_credentials.rs
```diff
@@ -0,0 +1,6 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+// TODO: would this better live in common?
```
cgettys-microsoft, line 6: TODO: nit clear to me if this is used outside of client

crates/libs/core/src/types/client/settings.rs
```diff
@@ -0,0 +1,539 @@
...
+/// Note: you can default fields you're not interested in like so:
+/// ```
+/// let my_settings = FabricClientSettings {
+///  PartitionLocationCacheLimit: Option<NonZeroU32>(NonZeroU32::new(1).expect("Non-zero value")),
+///  ..Default::default()
+/// };
+/// ```
+#[derive(Default, Clone)]
+#[allow(non_snake_case, reason = "For consistency with underlying COM api")]
+pub struct FabricClientSettings {
```
youyuanwu, line 45: This seems complicated.
I would just do the simplest code to handle EX like here:
https://github.com/Azure/service-fabric-rs/blob/883ca784d4493a5b54b94a1b7cd5977766e6bd69/crates/libs/core/src/client/query_client.rs#L139

cgettys-microsoft: Better?

PR #139, opened 2025-03-07
Start adding support for FabricClientSettings and FabricSecurityCredentials

cgettys-microsoft:
I'd like to have support for ergonomic (i.e. safe) versions of FABRIC_CLIENT_SETTINGS and FABRIC_SECURITY_CREDENTIALS
This makes necessary error handling changes (that are breaking), as FabricClientBuilder::build() now can error if there are invalid settings. It's a small step towards #140 
* For now, the error type remains opaque (as it isn't actually instantiated anyway).
* The FabricClientSettings struct remains empty (and adding new Optional defaults shouldn't breaking)
* The FabricSecurityCredentials enum is also empty (and marked non_exhaustive, so adding more entries in future will not be a breaking change)
* Also add some trivial wrapper types for the next PR, empty modules, and so on.


crates/libs/core/src/client/mod.rs
```diff
@@ -80,7 +93,24 @@ fn create_local_client_internal<T: Interface>(
...
-    .expect("failed to create fabric client")
+    .expect("failed to create fabric client");
+    if client_settings.is_some() || client_credentials.is_some() {
+        let setting_interface = client
+            .clone()
+            .cast::<IFabricClientSettings2>()
+            .expect("failed to cast fabric client to IFabricClientSettings2");
+        if let Some(desired_settings) = client_settings {
+            desired_settings
+                .apply(&setting_interface)
```
youyuanwu, line 104: IMO the combining of the setting value should be the responsibility of the user. U can provide api to load existing values.

cgettys-microsoft: Sure, but I'm not doing the combining of the setting value in this PR at all - it just is applying the settings, if provided. 
The "what does that mean, do we combine settings" isn't in this PR at all.

I could make two methods - with_client_settings_overrides and with_client_settings, if we want to expose  the capability to not get defaults from the client. But they can also always fall back to the COM API for that sort of advanced scenario?

Or are you saying you'd rather client settings and credentials not be part of the builder, but instead be &mut self methods on FabricClient and the individual clients as well?

crates/libs/core/src/client/mod.rs
```diff
@@ -34,13 +38,22 @@ pub use notification::ServiceNotification;
 #[cfg(test)]
 mod tests;
 
+#[non_exhaustive]
+#[derive(Debug)]
+pub enum FabricClientCreationError {
```
youyuanwu, line 43: Impl std Error trait?

crates/libs/core/src/client/mod.rs
youyuanwu: with_client_settings.
FabricClientSettings should have fn to load existing settings.

crates/libs/core/src/types/common/security_credentials/mod.rs
```diff
@@ -0,0 +1,27 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+#![deny(unsafe_op_in_unsafe_fn)]
```
youyuanwu, line 5: Where are denies needed?

youyuanwu (approved):

crates/libs/core/src/client/mod.rs
cgettys-microsoft: Will do in follow up PR

crates/libs/core/src/types/common/security_credentials/mod.rs
youyuanwu: I don't see unsafe code in this mod in this PR. Assme it will be used in the next.

PR #141, opened 2025-03-11
Add mssf_core support for Windows, Claims, and X509 Credentials

cgettys-microsoft:
* Add mssf_core support for Windows, Claims, and X509 Credentials
* Make error type added in #139 implement Error
* Add mock IFabricSettingsClient2 implementation to allow testing the new code; will be used more when FabricClientSettings PR is ready.
* Tests of the new code pass in Miri under Tree Borrows, but not under stack borrows. Details below:
```
// COM interface code  like this is not valid under stacked borrows, but appears to be fine under tree borrows
//  let this: &Identity = &*((this as *const *const ()).offset(OFFSET) as *const Identity);
// So we run with only tree-borrows (that's what -Zmiri-disable-stacked-borrows -Zmiri-tree-borrows does)
// -Zmiri-ignore-leaks is to work around WeakReference shenanigans in the windows_core interface Drop implementation; they appear to be UB even under tree borrows? Should follow up on this too. But given that the actual implementation is written in C++, those aren't concerning for real usage - so I'm simply leaking the IFabricClientSettings2 implementation when running under Miri to ensure that there are no other sources of UB.
$env:MIRIFLAGS="-Zmiri-backtrace=full -Zmiri-disable-stacked-borrows -Zmiri-tree-borrows -Zmiri-disable-isolation -Zmiri-ignore-leaks"
// Currently Windows miri emulation does not work because of GetErrorInfo calls that aren't supported presently; should probably see if we can upstream support.
cargo +nightly miri test security_credentials --target x86_64-unknown-linux-gnu
```
* Leaving some extensions and other Fabric Credentials kinds as future work.


crates/libs/core/src/types/common/security_credentials/claims_credentials.rs
```diff
@@ -4,4 +4,277 @@
...
+        let call_counter_copy = Arc::clone(&call_counter);
+        let com = MockIFabricClientSettings::new_with_security_credentials_mock(Box::new(
+            move |creds: *const FABRIC_SECURITY_CREDENTIALS| {
+                *call_counter_copy.lock().expect("Not poisoned") += 1;
+                assert!(!creds.is_null() && creds.is_aligned());
+                // SAFETY: test code. non-null and alignment is checked above
+                let creds_copy: FABRIC_SECURITY_CREDENTIALS = unsafe { ptr::read(creds) };
+                assert_eq!(creds_copy.Kind, FABRIC_SECURITY_CREDENTIAL_KIND_CLAIMS);
+
+                let value = creds_copy.Value as *const FABRIC_CLAIMS_CREDENTIALS;
```
youyuanwu, line 162: why not use value.as_ref().unwrap() to get a reference?

crates/libs/core/src/types/common/security_credentials/claims_credentials.rs
```diff
@@ -4,4 +4,277 @@
...
+    use std::ptr;
+    use std::sync::{Arc, Mutex};
+
+    use crate::strings::WStringWrap;
+    use crate::types::mockifabricclientsettings::test_constants::*;
+    use crate::types::mockifabricclientsettings::test_utilities::check_array_parameter;
+    use crate::types::mockifabricclientsettings::MockIFabricClientSettings;
+
+    use super::*;
+    fn make_credentials() -> FabricClaimsCredentials {
```
youyuanwu, line 102: If u implemented a type FabricClaimsCredentials  that can convert into and from FABRIC_CLAIMS_CREDENTIALS would make the testing easier. Mocking the COM set api makes things complicated.

youyuanwu: Is there a use case where user just want to get all the settings?

youyuanwu (approved):

crates/libs/core/src/types/common/security_credentials/claims_credentials.rs
cgettys-microsoft: There's no GetSecurityCredentials on the interface, afaik. So no.

cgettys-microsoft: And sure, but I don't think it'd really make things cleaner - just would add even more boilerplate for deserializing each variant of the COM struct.

crates/libs/core/src/types/common/security_credentials/claims_credentials.rs
youyuanwu: All other places in this crate uses as_ref() to access the c struct and ptr::read is not used.

cgettys-microsoft: Fixed

youyuanwu (approved):

PR #144, opened 2025-03-12
Remove ctrlc dependency

youyuanwu:
Tokio signal has ctrl-c integration, so no need to use another crate.
Users need to replace the run_until_ctrl_c() with this:
```rs
 tokio::signal::ctrl_c().await
```

### DIFF
--- a/crates/libs/core/src/runtime/executor.rs
+++ b/crates/libs/core/src/runtime/executor.rs
@@ -5,8 +5,7 @@
 
 use std::future::Future;
 
-use tokio::{runtime::Handle, sync::mpsc::channel};
-use tracing::info;
+use tokio::runtime::Handle;
 
 use crate::error::ErrorCode;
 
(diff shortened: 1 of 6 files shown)


OscarTHZhang (approved):

crates/libs/core/src/runtime/executor.rs
```diff
@@ -28,28 +27,6 @@ pub trait Executor: Clone + Sync + Send + 'static {
 
     /// run the future on the executor until completion.
     fn block_on<F: Future>(&self, future: F) -> F::Output;
-
-    // provided functions
-
-    /// Run the executor and block the current thread until ctrl-c event is
-    /// Received.
-    fn run_until_ctrl_c(&self) {
```
cgettys-microsoft, line 36: I'm not opposed to this, better to leave it in user's hands, but you could also replace the implementation with the provided one-liner, no?

youyuanwu: The trait should not depend on a specific runtime. Unless tokio's ctrl-c is cross runtime.

cgettys-microsoft: Ah, right, good point, no reason to add another dependency on details of the runtime.

PR #145, opened 2025-03-12
Refactor mssf_core::sync mod

youyuanwu:
Remove unused code and tests, including macro generated fabric client that never went anywhere.
Renamed some reimplementation of the sync types like BridgeContext3 to BridgeContext.
Moved code to appropriate files that was planned long ago.


cgettys-microsoft:
Very nice!

crates/libs/core/src/sync/channel.rs
```diff
@@ -62,46 +159,77 @@ impl<T> FabricSender<T> {
             // If it acctually dropped by user, it is ok to ignore because user
             // does not want to want the value any more. But too bad SF has done
             // the work to get the value.
-            debug_assert!(false, "receiver dropped.");
+
+            // receiver should never be dropped if operation is not cancelled.
+            if let Some(t) = self.token {
+                debug_assert!(
```
cgettys-microsoft, line 165: Consider a retail assert, unless we expect this to be performance critical?

youyuanwu: In rare cases or special apis, user may violate this unintentionally. I don't want to cause the process to terminate in prod, since this does not affect correctness, since the receiver has already given up.

PR #146, opened 2025-04-02
Replace debug trace with trace instrument macro

youyuanwu:
Added feature trace to enable tracing.
The bridge and proxy layer traces are instrumented by macro. The importance of the trace at bridge and proxy layer is to capture the return and error values. Users of mssf should do their own tracing with more verbosities.
Formatted all toml files.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,11 +10,13 @@ members = [
 
 [workspace.dependencies]
 tracing = { version = "0.1", features = ["log"] }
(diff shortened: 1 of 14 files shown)


qbradley (approved):
Will you revise the crate version with a future PR, or should they be changed with this one since you are adding a new feature?

youyuanwu:
> Will you revise the crate version with a future PR, or should they be changed with this one since you are adding a new feature?

Crate version will be updated in a separate PR before new release is created.

PR #149, opened 2025-04-21
Add get_deployed_replica_detail to fabric client

sahemant:
**Description**
Currently, The GetDeployedReplicaListAsync is exposed in Fabric Client API. This is required to find the read and write status of a Stateful Replica along with other fields.
This PR adds the `get_deployed_replica_detail` API to fabric query client. However, the list types are ignored for now as the memory owned by com object must be saved when using list with shared pointer. 
**Testing:**
Added unit test which invokes get_deployed_replica_detail and verifies stateful server parition write and read status


crates/libs/core/src/types/client/mod.rs
```diff
@@ -93,3 +104,59 @@ impl From<ClientRole> for FABRIC_CLIENT_ROLE {
         }
     }
 }
+
+// FABRIC_QUERY_SERVICE_OPERATION_NAME
+#[derive(Debug, Clone)]
```
youyuanwu, line 109: add PartialEq

crates/libs/core/src/types/client/metrics.rs
```diff
@@ -125,3 +126,28 @@ type PrimaryLoadMetricReportListIter<'a> =
     FabricIter<'a, FABRIC_LOAD_METRIC_REPORT, LoadMetricReport, PrimaryLoadMetricReportList>;
 type SecondaryLoadMetricReportListIter<'a> =
     FabricIter<'a, FABRIC_LOAD_METRIC_REPORT, LoadMetricReport, SecondaryLoadMetricReportList>;
+
+// FABRIC_LOAD_METRIC_REPORT_LIST
+#[derive(Debug, Clone)]
+pub struct LoadMetricReportList {
+    pub count: u32,
```
youyuanwu, line 133: List needs to have a com object owner as a field. otherwise items can have wrong lifetimes.

youyuanwu: Remove this for now if it is unused. Add comment for fixing the api in future.

crates/libs/core/src/types/client/mod.rs
sahemant: added PartialEq for both enums introduced in this PR.

crates/libs/core/src/types/client/metrics.rs
sahemant: Done. Removed this and `replicator.rs` (introducted in initial commit) which are unused. commented out reported_load and replicator_status fields with `//TODO` in `DeployedStatefulServiceReplicaDetailQueryResult` as both needs access to List.

youyuanwu (approved):

Issue #152, opened 2025-04-24
Support CreateService and UpdateService FabricClient bindings


youyuanwu:
Csharp corresponding api:
https://learn.microsoft.com/en-us/dotnet/api/system.fabric.fabricclient.servicemanagementclient.createserviceasync?view=azure-dotnet

https://learn.microsoft.com/en-us/dotnet/api/system.fabric.fabricclient.servicemanagementclient.updateserviceasync?view=azure-dotnet

youyuanwu:
Completed by #155 #154

PR #153, opened 2025-04-25
FabricClient: Add create delete service api

youyuanwu:
Support #152 
UpdateService will be implemented in a separate PR.
Added test for a sample app, create and delete services for various partition schemes.
Deleted the code in sample app to start listener, since it creates port conflict.

### DIFF
--- a/crates/libs/core/src/types/client/service.rs
+++ b/crates/libs/core/src/types/client/service.rs
@@ -3,6 +3,11 @@
 // Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
 // ------------------------------------------------------------
 
+#![cfg_attr(
+    not(feature = "tokio_async"),
+    allow(dead_code, reason = "code configured out")
+)]
+
 use std::ffi::c_void;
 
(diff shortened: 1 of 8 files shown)


crates/libs/core/src/types/client/service.rs
```diff
@@ -0,0 +1,250 @@
...
+use crate::types::{MoveCost, PartitionSchemeDescription, ServicePackageActivationMode, Uri};
+
+pub enum ServiceDescription {
+    // Invalid,
+    Stateful(StatefulServiceDescription), // FABRIC_STATEFUL_SERVICE_DESCRIPTION
+    Stateless(StatelessServiceDescription), // FABRIC_STATELESS_SERVICE_DESCRIPTION
+}
+
+#[derive(Debug, Default)]
+pub struct StatefulServiceDescription {
```
youyuanwu, line 33: Eventually I need to add a builder type. This struct is not friendly to create.

OscarTHZhang (approved):

jrahman (approved):

PR #154, opened 2025-04-29
FabricClient: add update service api

youyuanwu:
Csharp equivalent api: [updateserviceasync](https://learn.microsoft.com/en-us/dotnet/api/system.fabric.fabricclient.servicemanagementclient.updateserviceasync?view=azure-dotnet)

Only stateful services are supported for now.
Added test to add and remove named partitions dynamically.

### DIFF
--- a/crates/libs/core/src/types/client/service.rs
+++ b/crates/libs/core/src/types/client/service.rs
@@ -11,15 +11,23 @@
 use std::ffi::c_void;
 
 use mssf_com::FabricTypes::{
-    FABRIC_SERVICE_DESCRIPTION, FABRIC_SERVICE_DESCRIPTION_KIND_STATEFUL,
-    FABRIC_SERVICE_DESCRIPTION_KIND_STATELESS, FABRIC_STATEFUL_SERVICE_DESCRIPTION,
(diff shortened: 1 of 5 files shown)


crates/libs/core/src/types/client/service.rs
```diff
@@ -248,3 +263,351 @@ impl ServiceDescription {
...
+    pub names_to_remove: Vec<WString>,
+}
+
+/// Holder for memories passed to the FFI
+pub(crate) struct NamedRepartitionDescriptionRaw<'a> {
+    _names_to_add: Vec<PCWSTR>,
+    _names_to_remove: Vec<PCWSTR>,
+    internal: Box<FABRIC_NAMED_REPARTITION_DESCRIPTION>,
+    // owner of string buffers
+    _owner: &'a NamedRepartitionDescription,
```
OscarTHZhang, line 290: Can this be PhantomData?

OscarTHZhang (approved):
LGTM. Suggest adding some more documentation on generally how to handle raw pointer of fabric extensions for future reference

crates/libs/core/src/types/client/service.rs
youyuanwu: possibly. Will try in the next PR.

PR #155, opened 2025-05-02
FabricClient: Finalize create service api

youyuanwu:
Added builder pattern for create service api, removed pub struct fields.
Moved to use PhantomData when possible.

### DIFF
--- a/crates/libs/core/src/types/client/service.rs
+++ b/crates/libs/core/src/types/client/service.rs
@@ -8,7 +8,7 @@
     allow(dead_code, reason = "code configured out")
 )]
 
-use std::ffi::c_void;
+use std::{ffi::c_void, marker::PhantomData};
 
 use mssf_com::FabricTypes::{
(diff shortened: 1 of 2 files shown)


crates/libs/core/src/types/client/service.rs
```diff
@@ -133,43 +204,91 @@ impl StatefulServiceDescription {
...
+        self
+    }
+}
 pub(crate) struct StatelessServiceDescriptionRaw<'a> {
     internal: Box<FABRIC_STATELESS_SERVICE_DESCRIPTION>,
     _internal_ex1: Box<FABRIC_STATELESS_SERVICE_DESCRIPTION_EX1>,
     _internal_ex2: Box<FABRIC_STATELESS_SERVICE_DESCRIPTION_EX2>,
     _internal_ex3: Box<FABRIC_STATELESS_SERVICE_DESCRIPTION_EX3>,
     _internal_ex4: Box<FABRIC_STATELESS_SERVICE_DESCRIPTION_EX4>,
     // String buffers memory owner
```
jrahman, line 290: Is this comment still correct?

jrahman (approved):

PR #156, opened 2025-05-05
Support ubuntu devcontainer

youyuanwu:
Support ubuntu devcontainer.
Use docker compose based devcontainer, where 1 container runs a sf onebox cluster, and the other hosts the source code for development.
Added ci for devcontainer. It builds the devcontainer first, and then compiles code and run tests. We have test parity between windows and devcontainer.


crates/samples/echomain-stateful2/src/main.rs
```diff
@@ -14,7 +14,7 @@ use tracing::info;
 mod echo;
 mod statefulstore;
 // Disable test for Linux ci for now due to SF app problem
-#[cfg(target_os = "windows")]
+//#[cfg(target_os = "windows")]
```
qbradley, line 17: Maybe remove the commented code if it is not windows specific anymore.

qbradley: and update the comment on line 16

qbradley (approved):

PR #158, opened 2025-05-07
Prepare 0.0.23 release

youyuanwu:
Revert to use nightly for packaging test since the package-workspace feature is still unstable.

### DIFF
--- a/.github/workflows/generate.yaml
+++ b/.github/workflows/generate.yaml
@@ -41,5 +41,7 @@ jobs:
     - name: Install rust nightly
       uses: dtolnay/rust-toolchain@nightly
 
+    # TODO: move back to stable once package-workspace is stabilized
+    # https://doc.rust-lang.org/cargo/reference/unstable.html#package-workspace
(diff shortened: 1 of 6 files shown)


.github/workflows/generate.yaml
```diff
@@ -41,5 +41,7 @@ jobs:
     - name: Install rust nightly
       uses: dtolnay/rust-toolchain@nightly
 
+    # TODO: move back to stable once package-workspace is stabilized
+    # https://doc.rust-lang.org/cargo/reference/unstable.html#package-workspace
     - name: check crate packaging
```
jrahman, line 46: What is the expected timeline for stabilization? Not a big fan of relying on nightly here if we can avoid it

jrahman (approved):

.github/workflows/generate.yaml
youyuanwu: Not sure when it stablizes. But this in the past has caught some legit issues.

PR #159, opened 2025-06-04
[FabricClient] CreateService an UpdateService api Send+Sync

youyuanwu:
CreateService and UpdateService api and the parameter arguments should be Send+Sync in order to be used in runtime tasks.
Changed existing tests to call these apis on a separate task to validate the compilation.

### DIFF
--- a/crates/libs/core/src/types/common/partition.rs
+++ b/crates/libs/core/src/types/common/partition.rs
@@ -183,6 +183,10 @@ pub struct UniformIn64PartitionSchemeDescription {
     internal: Box<FABRIC_UNIFORM_INT64_RANGE_PARTITION_SCHEME_DESCRIPTION>,
 }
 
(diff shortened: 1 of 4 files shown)


crates/libs/core/src/types/common/partition.rs
```diff
@@ -218,6 +222,10 @@ pub struct NamedPartitionSchemeDescription {
     internal: Box<FABRIC_NAMED_PARTITION_SCHEME_DESCRIPTION>,
 }
 
+/// SAFETY: This is thread safe because the raw pointers points to heap allocated memory.
+unsafe impl Send for NamedPartitionSchemeDescription {}
+unsafe impl Sync for NamedPartitionSchemeDescription {}
```
jrahman, line 227: Technically `Sync` would be safe not because the pointer points at heap memory, but rather because the data is read-only, and does not need synchronization

jrahman (approved):

PR #163, opened 2025-06-19
FabricClient: Fabric Property Client CRUD APIs

youyuanwu:
Implement basic CRUD apis for Fabric Property Client.
Batch and other apis will be added in follow up PRs.
Added tests to create property for the echo app.
The corresponding csharp api is here: [PropertyManagementClient](https://learn.microsoft.com/en-us/dotnet/api/system.fabric.fabricclient.propertymanagementclient?view=azure-dotnet)

### DIFF
--- a/crates/libs/core/src/client/property_client.rs
+++ b/crates/libs/core/src/client/property_client.rs
@@ -1,0 +1,611 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+#![cfg_attr(
+    not(feature = "tokio_async"),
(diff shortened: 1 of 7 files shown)


crates/libs/core/src/client/property_client.rs
```diff
@@ -0,0 +1,639 @@
...
+                    timeout_milliseconds,
+                    callback,
+                )
+            },
+            move |ctx| unsafe { com2.EndGetProperty(ctx) },
+            cancellation_token,
+        )
+    }
+
+    // TODO: implement this
```
qbradley, line 322: Is this TODO and the following TODO still relevant?

qbradley: Or the comment is not to implement the code here, but to expose it in PropertyManagementClient ?

qbradley (approved):

PR #175, opened 2025-07-29
WIP: ServiceResolver

youyuanwu:


crates/libs/core/src/error/errorcode.rs
```diff
@@ -103,6 +103,28 @@ impl core::fmt::Display for ErrorCode {
     }
 }
 
+impl ErrorCode {
+    /// Matches the FabricTransientException mapping in InteropExceptionMap.cs
+    pub fn is_transient(&self) -> bool {
+        matches!(
```
OscarTHZhang, line 109: Is there an official list of retriable transient errors?

OscarTHZhang (approved):
LGTM

PR #185, opened 2025-08-19
mssf-util crate focusing on tokio integration

youyuanwu:
Add mssf-util new crate, where it focuses on tokio integration.
Tokio required code will be gradually moved from mssf-core into mssf-util (under tokio feature flag).
Apps built on tokio will be easier to write using mssf-util.
mssf-util will likely to expand to include more service resolution or failover detection feature.

### DIFF
--- a/crates/libs/util/src/tokio/mod.rs
+++ b/crates/libs/util/src/tokio/mod.rs
@@ -104,15 +104,20 @@ impl<T: Send> JoinHandle<T> for TokioJoinHandle<T> {
     }
 }
 
+/// Sleep timer implementation for tokio
 pub struct TokioTimer;
 
+// TODO: the return type may be simplified if using return impl
 impl Timer for TokioTimer {
     fn sleep(&self, duration: std::time::Duration) -> std::pin::Pin<Box<dyn Sleep>> {
         Box::pin(TokioSleep::new(tokio::time::sleep(duration)))
     }
 }
 
+/// Sleep future implementation for tokio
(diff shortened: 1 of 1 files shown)


crates/libs/util/src/tokio/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+                };
+                #[cfg(feature = "tracing")]
+                tracing::error!("DefaultJoinHandle: background task failed: {ec}, msg: {e}");
+                Err(ec.into())
+            }
+        }
+    }
+}
+
+pub struct TokioTimer;
```
qbradley, line 107: Can you add documentation for TokioTimer and TokioSleep? What are they for and why are the implemented this way?

Why can't TokioTimer::sleep return a TokioSleep instead of Pin<Box<dyn Sleep>>?
Why can't TokioSleep contain inner as type tokio::time::Sleep instead of Pin<Box<tokio::time::sleep>> since the Future implementation gets self as a Pin<&mut Self>?

qbradley (approved):

crates/libs/util/src/tokio/mod.rs
youyuanwu: 1. The trait return type is currently Pin<Box>. If we change to return impl for the trait, the Pin<Box> could be removed.
2. The inner tokio::time::sleep cannot unpin due to some phantom data. May need pin_project macro to do it. 

They are future considerations for improvements.

PR #195, opened 2025-09-17
FabricClient: Add cluster health api

youyuanwu:
Also add cluster health in monitoring api.

Added new mechanism to construct COM structs using BoxPool:  COM api call ususally involves using COM structs that are short lived, and such structs have complex pointer associations. Previously different tricks were used to construct these structs and it is hard to maintain. With the introduction of BoxPool, COM structs can be easily allocated with raw pointers.
BoxPool needs to be kept in scope until the COM api call returns. This works because COM api always copies the struct into internal cpp structs before returning.
COM api used to return a iterator for com lists and it is quite complex to write. We plan to just return Vec<_> instead of iterator so that we don't need to manage lifetimes and boilerplate code. User usually convert the iterator to Vec immediately, so there is no point to support the iterator.

### DIFF
--- a/crates/libs/core/src/mem.rs
+++ b/crates/libs/core/src/mem.rs
@@ -1,0 +1,68 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+/// A pool to hold boxes to extend their lifetime.
+/// This is typically useful to build raw COM structures that require
+/// raw pointers in struct fields.
+#[derive(Debug, Default)]
+pub struct BoxPool {
+    inner: Vec<Box<dyn 'static + std::any::Any>>,
+}
+
+impl BoxPool {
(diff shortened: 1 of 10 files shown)


crates/libs/core/src/mem.rs
```diff
@@ -0,0 +1,68 @@
...
+        raw
+    }
+
+    /// Push a Vec<T> into the pool, and return its length and a raw pointer to its data.
+    /// The pointer is valid as long as the pool is alive.
+    pub fn push_vec<T: 'static>(&mut self, v: Vec<T>) -> (usize, *const T) {
+        let len = v.len();
+        let raw = v.as_ptr();
+        // Convert Vec<T> to Box<dyn Any>.
+        let boxed_v = Box::new(v);
```
OscarTHZhang, line 33: Does this involve extra memory allocation or is it complete transparent?

youyuanwu: This requires allocation. I tried with Box<[T]> there is problem with Sized and [T], push a boxed vector is good enough for now.

OscarTHZhang (approved):

PR #204, opened 2025-10-01
Remove u20 support

youyuanwu:
u20 is deprecated. Remove dev container and all ci tests for it.


OscarTHZhang (approved):

youyuanwu:
> Should we add Azure Linux support in the future?

It is in another branch already. Will publish PR soon.

PR #253, opened 2026-02-10
[core] Support cancellation token callback

youyuanwu:
User apps may use tokio_util::CancellationToken and mssf-core BoxedCancelToken is hard to link to it. In practice this difference forces the user to invent some logic to propagate the BoxedCancelToken signal to the tokio::CancellationToken using a backgound task.
This PR adds to the BoxedCancelToken to be able to inject an on_cancel callback for idiomatically propagate cancellation.
User can call tokio CancellationToken inside the callback.

### DIFF
--- a/crates/libs/core/src/runtime/executor.rs
+++ b/crates/libs/core/src/runtime/executor.rs
@@ -46,7 +46,7 @@ pub trait CancelToken: Send + Sync + 'static {
     /// Register a callback to be invoked when this token is cancelled.
     /// If this token is already cancelled, the callback is invoked immediately.
     ///
-    /// # Panics
+    /// # Panics (debug builds only)
     /// Panics if a callback has already been registered.
     fn on_cancel(&self, callback: Box<dyn FnOnce() + Send + Sync>);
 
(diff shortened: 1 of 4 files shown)


crates/libs/core/src/runtime/executor.rs
```diff
@@ -43,6 +43,19 @@ pub trait CancelToken: Send + Sync + 'static {
...
+    /// The child can be any `BoxedCancelToken` (i.e. any impl type).
+    /// If this token is already cancelled, the child is cancelled immediately.
+    ///
+    /// # Panics
+    /// Panics if a child has already been attached.
+    ///
+    /// **Note:** Circular attachment (e.g. self-attach or A→B→A) creates
+    /// reference cycles that will leak memory if the token is never cancelled.
+    /// Cancellation breaks the cycle by taking the child.
+    fn attach_child(&self, child: BoxedCancelToken);
```
foreverzet: Maybe add callback 'on_cancel'?
A little bit more flexible

chenwei75 (approved):

PR #256, opened 2026-02-12
[Aux] Change sample application to have 1 auxiliary replica

youyuanwu:
Changed local onebox to enable Auxiliary replica support. Added script for using custom cluster config for windows, and custom cluster start and stop scripts.
Minor changes for auxiliary replics:
* Added all error code in SF 11 in the mssf-core error list.
* Added auxiliary role types in replica roles.
* Added axuiliary types for service upgrade API.

A separate PR will be added to change the service creation API to support auxiliary replicas.
The sample app auxiliary is specified in application manifest default services.

image

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -157,11 +157,11 @@ jobs:
 
       - name: Configure Docker to use /mnt directory
         run: |
-          # Create Docker data directory in /mnt
-          sudo mkdir -p /mnt/docker
+          # Stop both docker service and socket to prevent auto-restart
+          sudo systemctl stop docker docker.socket
           
-          # Stop Docker service
-          sudo systemctl stop docker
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -144,6 +144,11 @@ jobs:
         rust-version: ${{ fromJSON(needs.setup.outputs.rust-versions) }}
     runs-on: ubuntu-latest
     steps:
+      # There is only 20G of free os disk space by default.
+      - name: Free OS disk space
+        run: |
+          sudo rm -rf /usr/local/lib/android /usr/share/dotnet /opt/ghc
```
youyuanwu: github is shrinking OS disk from 100GB+ to 70GB. So sometimes we run out of disk. Out of disk is observed on main branch.

PR #266, opened 2026-02-24
[FabricClient] Use BoxPool for create_service api

youyuanwu:
Migrate create_service api to use the BoxPool ffi infrastructure, removing much of the manual Box handling code.
This simplifies the ffi code. All ffi code will be migrated to use BoxPool eventually.

Added a MD guide file to describe the BoxPool usage.

### DIFF
--- a/crates/libs/core/src/client/svc_mgmt_client.rs
+++ b/crates/libs/core/src/client/svc_mgmt_client.rs
@@ -5,7 +5,12 @@
 
 use std::{ffi::c_void, time::Duration};
 
-use crate::{PCWSTR, WString, runtime::executor::BoxedCancelToken, types::Uri};
+use crate::{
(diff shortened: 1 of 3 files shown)


crates/libs/core/src/client/svc_mgmt_client.rs
```diff
@@ -335,8 +340,8 @@ impl ServiceManagementClient {
         cancellation_token: Option<BoxedCancelToken>,
     ) -> crate::Result<()> {
         {
-            let desc_raw = desc.build_raw();
-            let ffi_raw = desc_raw.as_ffi();
+            let mut pool = BoxPool::new();
```
chenwei75: What is the benefit of changing to BoxPool?

qbradley (approved):

PR #276, opened 2026-04-03
Use spawn_blocking for IFabricAsyncOperationCallback::Invoke

qbradley:
IFabricAsyncOperationCallback::Invoke can take locks and they are not always released quickly. Therefore, we do not want to invoke it directly from tokio I/O threads. spawn_blocking is used to move the callback invocation to a tokio thread_pool thread so that I/O threads (a scarce resource) remain available.

### DIFF
--- a/crates/libs/util/src/tokio/mod.rs
+++ b/crates/libs/util/src/tokio/mod.rs
@@ -68,6 +68,14 @@ impl Executor for TokioExecutor {
     {
         self.rt.spawn(future);
     }
+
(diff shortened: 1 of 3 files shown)


youyuanwu:
Addresses this #277

crates/libs/util/src/tokio/mod.rs
youyuanwu, line 28: We effectively only support tokio multi thread runtime. Need to add a check to deny current runtime somehow.

youyuanwu (approved):

PR #282, opened 2026-04-06
Cancel the SF operation on FabricReceiver Drop

### LINKED ISSUE #281: Fabric Sender panic
See run https://github.com/Azure/service-fabric-rs/actions/runs/24022077724/job/70052867379
```
thread 'tokio-rt-worker' (4678) panicked at crates/libs/core/src/sync/channel.rs:174:17:
task should be cancelled when receiver dropped.

thread 'tokio-rt-worker' (4678) panicked at /rustc/e408947bfd200af42db322daf0fadfe7e26d3bd1/library/core/src/panicking.rs:225:5:
panic in a function that cannot unwind
stack backtrace:
```

The problem is that Receiver is not properly cancelled, and we wrongly assumed that caller cannot drop the receiver.

youyuanwu:
Fixes: #281 
It was assumed that user never drops FabricReceiver without cancelling the token. However, this can happen in valid tokio::select branch.
The fix is to remove the send error assert, since there is no correctness problem. This change also auto cancels internal SF operations that is not completed and token not cancelled.

### DIFF
--- a/crates/libs/core/src/sync/channel.rs
+++ b/crates/libs/core/src/sync/channel.rs
@@ -167,7 +167,7 @@ impl<T> Drop for FabricReceiver<T> {
     fn drop(&mut self) {
         // Note that there is a chance that SF operation is not cancelled when user cancelled the token
         // and drop the receiver immediately without polling the receiver future.
-        // because the cancellation signal is only propergated when polling the receiver future.
+        // because the cancellation signal is only propagated when polling the receiver future.
         // However this very uncommon and violate the best practice that user should always poll
         // the receiver future to completion even after the cancellation is triggered.
         // SF operation will be left in the background and eventually finish.
--- a/crates/libs/core/src/sync/channel.rs
+++ b/crates/libs/core/src/sync/channel.rs
@@ -19,14 +19,14 @@ use crate::{
 pub use futures_channel::oneshot::{self, Receiver, Sender};
 
 // Token that wraps oneshot receiver.
-// The future recieve does not have error. This is designed for the use
+// The future receive does not have error. This is designed for the use
 // case where SF guarantees that sender will be called.
 pub struct FabricReceiver<T> {
     rx: Receiver<T>,
     token: Option<BoxedCancelToken>,
-    // event for cancelling
+    // event from the token, this is needed to poll the cancellation in the receiver future.
     cancel_event: Option<Pin<Box<dyn EventFuture + 'static>>>,
-    // saved ctx from SF Begin COM api for cancalling.
+    // saved ctx from SF Begin COM api for cancelling.
     ctx: Option<IFabricAsyncOperationContext>,
 }
 
@@ -74,6 +74,12 @@ impl<T> FabricReceiver<T> {
         }
         Ok(())
     }
+
+    // Cancel token no longer needed.
+    fn clear_cancel_fields(&mut self) {
+        self.token.take();
+        self.cancel_event.take();
+    }
 }
 
 // Returns error if cancelled.
@@ -100,12 +106,16 @@ impl<T> Future for FabricReceiver<T> {
         // Try to receive the value from the sender
         let inner = <Receiver<T> as Future>::poll(Pin::new(&mut this.rx), cx);
         match (inner, this.token.as_ref()) {
-            (Poll::Ready(Ok(data)), _) => Poll::Ready(Ok(data)),
+            (Poll::Ready(Ok(data)), _) => {
+                // Result received successfully; clear the token so Drop
+                // does not cancel the operation.
+                this.clear_cancel_fields();
+                Poll::Ready(Ok(data))
+            }
             (Poll::Ready(Err(_)), Some(t)) => {
                 if t.is_cancelled() {
                     // clear the token since we only propergate the signal once.
-                    this.token.take();
-                    this.cancel_event.take();
+                    this.clear_cancel_fields();
                     // cancel the SF ctx and clear it.
                     if let Err(e) = this.cancel_inner_ctx() {
                         Poll::Ready(Err(e))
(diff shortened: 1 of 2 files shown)


crates/libs/util/src/tokio/tests.rs
```diff
@@ -773,26 +773,35 @@ mod cancel_token_tests {
...
+            let err = result.expect_err("expected panic");
+            let msg = err
+                .downcast_ref::<&str>()
+                .copied()
+                .or_else(|| err.downcast_ref::<String>().map(|s| s.as_str()))
+                .unwrap_or("");
+            assert!(
+                msg.contains("a callback has already been registered"),
+                "unexpected panic message: {msg}"
+            );
```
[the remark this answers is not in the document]
youyuanwu: Reverted the change

crates/libs/core/src/sync/channel.rs
```diff
@@ -151,6 +160,21 @@ impl<T> Future for FabricReceiver<T> {
...
+// Cancel the token when the receiver is dropped without having received a value.
+// If nobody is waiting for the result, the operation should be cancelled.
+// This means that SF operation will be auto cancelled if user drops the receiver without waiting for the result.
+impl<T> Drop for FabricReceiver<T> {
+    fn drop(&mut self) {
+        if let Some(t) = self.token.take()
+            && !t.is_cancelled()
+        {
+            t.cancel();
+            // propergate to inner SF ctx, this is needed since rx is no longer polled.
```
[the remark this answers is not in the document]
youyuanwu: Changed to use on_cancel from token to propagate

youyuanwu:
Reverted to the first approach.

crates/libs/core/src/sync/channel.rs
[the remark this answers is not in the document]
youyuanwu: Reverted back. And documented this is intentional.

crates/libs/core/src/sync/channel.rs
```diff
@@ -164,14 +194,9 @@ impl<T> FabricSender<T> {
...
-            // the work to get the value.
-
-            // receiver should never be dropped if operation is not cancelled.
+            // Receiver's Drop impl cancels the token, so this should always hold.
             if let Some(t) = self.token {
-                debug_assert!(
+                assert!(
                     t.is_cancelled(),
                     "task should be cancelled when receiver dropped."
                 );
```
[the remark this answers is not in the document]
youyuanwu: This is intentional. The sender and receiver has this contract.

crates/libs/core/src/sync/proxy.rs
```diff
@@ -29,13 +29,16 @@ use super::{FabricReceiver, oneshot_channel};
...
+/// Cancelling the token will in turn cancel the fabric operation.
 /// After cancellation is triggered, the receiver future should finish in a short time,
-/// with an error code opeartion cancelled, or other code if cancel failed.
+/// with an error code operation cancelled, or other code if cancel failed.
 /// If the result is ready before the cancellation is triggered, the success result will
 /// be the output of the receiver future.
+///
+/// Cancellation best practice:
+/// User should always poll the receiver future to completion even after cancellation is triggered,
+/// to ensure the cancellation signal is properly propagated to SF and resources are cleaned up in a timely manner.
```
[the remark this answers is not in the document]
youyuanwu: PR description fixed.

crates/libs/core/src/sync/channel.rs
```diff
@@ -19,14 +19,14 @@ use crate::{
 pub use futures_channel::oneshot::{self, Receiver, Sender};
 
 // Token that wraps oneshot receiver.
-// The future recieve does not have error. This is designed for the use
+// The future receive does not have error. This is designed for the use
```
jrahman: future receiver?

crates/libs/core/src/sync/channel.rs
```diff
@@ -19,14 +19,14 @@ use crate::{
...
 // case where SF guarantees that sender will be called.
```
jrahman: Does the new Drop impl invalidate this?

crates/libs/core/src/sync/channel.rs
```diff
@@ -100,12 +106,16 @@ impl<T> Future for FabricReceiver<T> {
...
-            (Poll::Ready(Ok(data)), _) => Poll::Ready(Ok(data)),
+            (Poll::Ready(Ok(data)), _) => {
+                // Result received successfully; clear the token so Drop
+                // does not cancel the operation.
+                this.clear_cancel_fields();
+                Poll::Ready(Ok(data))
+            }
             (Poll::Ready(Err(_)), Some(t)) => {
                 if t.is_cancelled() {
                     // clear the token since we only propergate the signal once.
```
qbradley: s/propergate/propagate/

qbradley: and a few other places in the file

qbradley (approved):
s/propergate/propagate/

crates/libs/core/src/sync/channel.rs
```diff
@@ -151,6 +160,27 @@ impl<T> Future for FabricReceiver<T> {
...
+        // Note that there is a chance that SF operation is not cancelled when user cancelled the token
+        // and drop the receiver immediately without polling the receiver future.
+        // because the cancellation signal is only propagated when polling the receiver future.
+        // However this very uncommon and violate the best practice that user should always poll
+        // the receiver future to completion even after the cancellation is triggered.
+        // SF operation will be left in the background and eventually finish.
+        if let Some(t) = self.token.take()
+            && !t.is_cancelled()
+        {
+            t.cancel();
```
jrahman: It makes sense why cancel_inner_ctx() is called, but isn't `t` the user supplied cancellation token from further up the stack? Couldn't that cancellation token be shared with other pending futures that would be cancelled as a side-effect of this future being dropped, do we want that?

youyuanwu: Good point. I may need to create a new token that is chained to the user's token, but only cancel this new token,

crates/libs/core/src/sync/channel.rs
youyuanwu: Updated the comments

crates/libs/core/src/sync/channel.rs
youyuanwu: Changed to never cancel user token.

PR #285, opened 2026-04-10
[test] Trace error on query partition error and increase retry

youyuanwu:
Changed reflection app tests to use tracing instead of print.
Changed the query partition to return empty vec result on failure for exposing the SF bug that repartition gets stuck. Also increase the max retry to make this stuck apparent.


youyuanwu:
This run hit the repartition issue: https://github.com/Azure/service-fabric-rs/actions/runs/24257831037/job/70834022929

OscarTHZhang (approved):

PR #292, opened 2026-04-30
mssf_util::tonic — failover-aware tonic Channel for Service Fabric

youyuanwu:
This is part of the work for proxyless service mesh on top of SF naming resolution.
Ships a tonic-compatible channel that auto-resolves SF endpoints, reconnects after failover, and reacts to a server-signalled 'mssf-status' trailer. Gated by the 'tonic' cargo feature.

Public surface (re-exported flat from mssf_util::tonic):
  - TargetResolver trait + BoxError
  - DialTarget, TargetSelector, SelectError
  - FabricTargetResolver(+Builder)  —> SF naming impl
  - TargetConnector(+Builder)        —> Service<Uri>: resolve + TCP dial
  - SwapChannel                      —> ArcSwap<tonic::Channel>
  - ResolveStatusMiddleware       —> trailer dedup + rebuild trigger
  - TargetChannel + TargetChannelBuilder —> convenience composition

Architecture: 4 composable layers (middleware over SwapChannel over tonic::Channel over TargetConnector over TargetResolver). Trailer dedup state machine handles storm, reset, dumb-server escape hatch. No background task; no in-flight kill; complaint-based resolve only.

Tests:
  - 6 dedup classify() unit tests
  - 7 middleware integration tests (scripted Service + bodies)
  - 3 e2e failover tests in new mssf-tests crate using two real HTTP/2 servers + switchable resolver + graceful shutdown

Design doc: docs/design/TonicConnectorDesign.md.
TLS adapter and live-cluster sample test deferred to Future Work.

Part of #293

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
+name = "arc-swap"
+version = "1.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a3a1fd6f75306b68087b831f025c712524bcb19aad54e557b1129cfa0a2b207"
+dependencies = [
+ "rustversion",
+]
+
+[[package]]
 name = "async-trait"
 version = "0.1.89"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -217,12 +226,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "futures"
+version = "0.3.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b147ee9d1f6d097cef9ce628cd2ee62288d963e16fb287bd9286455b241382d"
+dependencies = [
+ "futures-channel",
+ "futures-core",
+ "futures-io",
+ "futures-sink",
+ "futures-task",
+ "futures-util",
+]
+
+[[package]]
 name = "futures-channel"
 version = "0.3.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "07bbe89c50d7a535e539b8c17bc0b49bdb77747034daa8087407d655f3f7cc1d"
 dependencies = [
  "futures-core",
+ "futures-sink",
 ]
 
 [[package]]
@@ -230,6 +254,12 @@ name = "futures-core"
 version = "0.3.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7e3450815272ef58cec6d564423f6e755e25379b217b0bc688e295ba24df6b1d"
+
+[[package]]
+name = "futures-io"
+version = "0.3.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cecba35d7ad927e23624b22ad55235f2239cfa44fd10428eecbeba6d6a717718"
 
 [[package]]
 name = "futures-macro"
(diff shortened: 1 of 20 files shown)


chenwei75:
Let copilot review this PR, the comments are below,
Review: changes requested
Thanks for the PR — the design is thoughtful, the layering is clean, and the added HTTP/2 failover tests are valuable. I think this is close, but I’d like to see a few issues addressed before merge.

1. with_tls API does not appear usable with real TLS transports
TargetChannelBuilder::with_tls is documented as accepting a TLS layer, but the bounds require the returned service to produce TokioIo<TcpStream>:

https://github.com/Azure/service-fabric-rs/blob/c259e7b271ea147ba32cb3267701cf1595079cd7/crates/libs/util/src/tonic/channel/builder.rs#L81-L89

SwapChannel also erases connectors to exactly:

https://github.com/Azure/service-fabric-rs/blob/c259e7b271ea147ba32cb3267701cf1595079cd7/crates/libs/util/src/tonic/channel/swap.rs#L21-L22

A real TLS connector generally returns a TLS-wrapped IO type, not plain TokioIo<TcpStream>, so this public API seems misleading or unusable for the documented purpose.

Suggested fix: either defer/remove with_tls until the TLS adapter lands, or generalize the connector erasure over an IO trait object/type that can represent TLS-wrapped streams.

2. Trailer dedup is not concurrency-safe
The dedup state transition does a load_full(), then later store() and rebuild():

https://github.com/Azure/service-fabric-rs/blob/c259e7b271ea147ba32cb3267701cf1595079cd7/crates/libs/util/src/tonic/middleware.rs#L173-L187

If multiple in-flight RPCs complete concurrently with the same trailer value, they can all observe the same previous value and all trigger rebuild(). That weakens the “storm-safe dedup” guarantee.

Suggested fix: serialize the classify/store/rebuild decision with a small mutex, or use a compare-and-swap loop so only one observer of a new value fires the rebuild.

3. FabricTargetResolver may replace cached RSP with an older one
This code replaces the cached RSP for any non-zero compare_version result:

https://github.com/Azure/service-fabric-rs/blob/c259e7b271ea147ba32cb3267701cf1595079cd7/crates/libs/util/src/tonic/naming/default.rs#L51-L56

Given the existing compare_version semantics, a positive result means the cached prev is newer than new_rsp. In that case, replacing the cache with new_rsp can move the resolver backward.

Suggested fix: only replace the cache when new_rsp is newer; keep prev when prev is same or newer.

4. Workspace dependency enables mssf-util/tonic globally
The root workspace dependency now enables the tonic feature:

https://github.com/Azure/service-fabric-rs/blob/c259e7b271ea147ba32cb3267701cf1595079cd7/Cargo.toml#L66-L70

That weakens the feature-gating story because workspace consumers inheriting mssf-util.workspace = true may pull in tonic/hyper/tower unintentionally.

Suggested fix: keep the workspace dependency feature-neutral and enable features = ["tonic"] only where needed, e.g. in tests/mssf-tests.

Overall: the architecture and test coverage look good, but I’d request changes for the TLS API mismatch and concurrent dedup behavior before merging.

youyuanwu:
Addressed PR review feedback
    
    - Remove broken with_tls API and reframe TLS as Future Work.
      TargetChannelBuilder::with_tls had bounds (Response = TokioIo<TcpStream>)
      that no real TLS connector could satisfy. Removed; SwapChannel docstring
      updated to call out the IO-type constraint; design doc gets a 'TLS
      (deferred)' section enumerating what enabling TLS will require.
    
    - Make trailer dedup concurrency-safe.
      Replace ArcSwapOption<String> with Mutex<Option<String>> so the
      load -> classify -> store sequence is atomic. Concurrent in-flight
      RPCs completing with the same trailer value now collapse to one
      rebuild() call. The mutex is released before invoking rebuild() so
      back-to-back rebuilds don't serialize. Adds regression test
      concurrent_same_value_dedups (32 concurrent dispatches, asserts 1
      rebuild).
    
    - Don't replace cached RSP with an older one.
      FabricTargetResolver previously matched 'Ok(_)' (catching both
      same-version and older replies) and replaced the cache, which could
      move the resolver backward. Switch to PartialOrd, only advance on
      Some(Ordering::Less); keep prev for equal/newer; reset on None
      (cross-service).
    
    - Restore feature-gating discipline.
      Workspace mssf-util dep is feature-neutral again. Only mssf-tests
      opts in to features = ['tonic']. Sample crates and downstream
      consumers no longer pull tonic/hyper/tower transitively.

chenwei75 (approved):

PR #296, opened 2026-05-06
[test]: replica lifecycle state machine + fail-lifecycle tests

youyuanwu:
- New crate::lifecycle module (Lifecycle struct + LifecycleState enum) enforcing the IStatefulServiceReplica contract with five states: Created -> Opening -> Active -> Closing -> Terminal. Lifecycle methods log a structured tracing::error! and return E_UNEXPECTED on contract violation. 19 unit tests cover every transition and rejection path.

- Replica's open/close lifecycle methods now use two-phase transitions (enter_opening + complete_open, enter_closing + complete_close), so a Decision::Fail leaves the state in Opening/Closing rather than prematurely advancing. This makes a follow-up SF abort observable on the wire, which fail_close_during_delete_then_abort relies on.

- Renamed tests/fail_change_role.rs -> tests/fail_lifecycle.rs and expanded it to cover three failure paths with strict assertions:
    * fail_open_then_approve_retry           SF reopens, no abort
    * fail_change_role_then_approve_retry    SF aborts + reopens
    * fail_close_during_delete_then_abort    SF aborts, no reopen

- ReplicaRegistry now replaces existing entries with the same (partition_id, replica_id) instead of pushing duplicates. SF's recovery activation after fail-open reuses the replica_id without abort, which previously left a stale controller entry that broke ListPending / Approve dispatch.

- test_cluster: process-wide FabricClient cache (fabric_client(), fabric_client_clone(), fabric_client_endpoint()) so tests don't pay the COM build cost per run. fail_lifecycle migrated to it.

- .cargo/config.toml: RUST_TEST_THREADS=1 to serialize tests within a binary while keeping cross-binary parallelism. Onebox-shared tests no longer race against each other inside the same process.

- mssf_core::ErrorCode::E_UNEXPECTED / E_FAIL replace ad-hoc HRESULT literals in statefulstore::Lifecycle::reject and grpc_control's Approve(FailMessage) handler.


crates/samples/reflection/src/lifecycle.rs
```diff
@@ -0,0 +1,430 @@
...
+        *s = LifecycleState::Terminal;
+        if prev == LifecycleState::Terminal {
+            tracing::warn!(
+                partition = ?self.partition_id,
+                replica = self.replica_id,
+                "abort called on already-terminal Replica; ignoring"
+            );
+            true
+        } else {
+            false
```
chenwei75: The return value of the abort is little odd, a duplicate call return true and a real abort return false?

.cargo/config.toml
```diff
@@ -0,0 +1,12 @@
...
+# Integration tests under crates/samples/reflection/tests/ talk to a
+# single shared onebox SF cluster. Running them in parallel within the
+# same test binary leads to gate-routing races and onebox load spikes,
+# so libtest is told to use a single thread per process.
+#
+# This affects layer 2 only (within one test binary). Layer 1 (running
+# multiple test binaries concurrently) is still parallel; we'll add
+# cargo-nextest if/when that becomes a problem.
+[env]
+RUST_TEST_THREADS = "1"
```
chenwei75, line 12: Do we want to put the single thread in workspace root? Possible later test support multiple threads?

chenwei75 (approved):
Some minor comments.

.cargo/config.toml
youyuanwu: This limits inside each test exe to run 1 test at a time. We have many test exe so there is still parallelism. 
We will switch to cargo nextest which supports better control. 
Currently after this config change there is no test time increase in the ci.

crates/samples/reflection/src/lifecycle.rs
youyuanwu: Changed to use enum.

PR #299, opened 2026-05-20
feat: tonic connector failover middleware + reflection app adoption

youyuanwu:
- Extend ResolveStatusMiddleware to inspect both initial response headers and trailer frames for the mssf-status signal, firing at most once
- Add tracing (info/warn) to FabricTargetResolver and middleware rebuild
- Implement Write RPC in reflection sample with access-status gating
- Add live-cluster failover integration test (tonic_failover)
- Add tonic_server_trailers wire-shape verification tests
- Consolidate ReflectionTonicConnectorAdoption.md into TonicConnectorDesign.md

### DIFF
--- a/crates/samples/reflection/tests/tonic_failover.rs
+++ b/crates/samples/reflection/tests/tonic_failover.rs
@@ -1,0 +1,214 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+//! Live-cluster integration test for the failover-aware tonic
+//! channel built by
+//! [`samples_reflection::grpc::build_primary_channel`].
+//!
+//! Drives the `Write` RPC against `fabric:/ReflectionApp/TonicFailoverTest`
+//! through a [`TargetChannel`](mssf_util::tonic::TargetChannel)
+//! and asserts that the channel rebuilds (and the next request
+//! succeeds against the new primary) after a `restart_replica`
+//! triggered failover. Companion to
+//! [`failover.rs`](failover.rs) — that test exercises the
+//! bare-resolve path; this one exercises the channel.
+//!
+//! Requires the `ReflectionApp` application package to be
+//! deployed on the local onebox (see top-level test
+//! instructions). The new partitioned service is created and
+//! deleted by the test itself.
+
+use std::time::Duration;
+
+use mssf_core::{
+    GUID, WString,
+    client::{FabricClient, svc_mgmt_client::PartitionKeyType},
+    types::Uri,
+};
+use mssf_util::resolve::ServicePartitionResolver;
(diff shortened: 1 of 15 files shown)


crates/samples/reflection/tests/tonic_failover.rs
```diff
@@ -0,0 +1,214 @@
...
+/// name in [`failover.rs`](failover.rs).
+async fn fabric_client_drop_hack(fc: FabricClient) {
+    tokio::time::sleep(Duration::from_secs(5)).await;
+    drop(fc);
+    tokio::time::sleep(Duration::from_secs(1)).await;
+}
+
+/// Resolve the test service until it has the full target
+/// replica set, then return the singleton partition id.
+async fn wait_for_ready(fc: &FabricClient, uri: &Uri) -> GUID {
```
qbradley, line 57: Probably should have a timeout or deadline on this in case service never resolves three endpoints.

crates/samples/reflection/tests/tonic_failover.rs
```diff
@@ -0,0 +1,214 @@
...
+            .await
+        }));
+    }
+    for h in handles {
+        let ack = h.await.unwrap();
+        tracing::info!(%ack, "concurrent ack");
+    }
+
+    // Teardown.
+    sm.delete_service(&uri).await;
```
qbradley, line 235: If this test fails, does it leak the service, and does that cause problems for subsequent tests?  Could we delete the service even on test failure, or do cleanup at the beginning of the test?

crates/samples/reflection/tests/tonic_failover.rs
```diff
@@ -0,0 +1,214 @@
...
+    for attempt in 0..max_attempts {
+        match client.write(write_request(partition_id, payload)).await {
+            Ok(resp) => {
+                let acked_by = resp.into_inner().acked_by;
+                tracing::info!(attempt, %acked_by, "write succeeded");
+                return acked_by;
+            }
+            Err(status) => {
+                tracing::info!(attempt, ?status, "write returned error; retrying");
+                last_err = Some(status);
```
qbradley, line 130: This is retrying on any error, not just Unavailable and transport errors per the comment above.  Maybe makes sense to have an allow list of errors and fail fast on the others?

qbradley (approved):

crates/samples/reflection/tests/tonic_failover.rs
youyuanwu: Added a timeout.

crates/samples/reflection/tests/tonic_failover.rs
youyuanwu: Added a cleanup at the beginning of the test.

crates/samples/reflection/tests/tonic_failover.rs
youyuanwu: Filtered retry on Unavailable and Unknown.

PR #303, opened 2026-06-04
[core] Support SF Last Error Message

youyuanwu:
SF sets the thread local error message on SF API call error, and user can use FabricGetLastErrorMessage to get the message.
In dotnet, all SF calls COM api calls retrieves the error message via FabricGetLastErrorMessage.
We implement the same logic here in mssf following dotnet: mssf_core::Error type adds an error message field and is populated after COM api call. This PR only enables this for all async Begin/End api pair, and future PR will enable all apis to capture last error message.

For example:
```rust
            let err = property_client
                .create_name(&Uri::from("fabric:/bad?x=1"), timeout, None)
                .await
                .unwrap_err();
            assert_eq!(
                err.to_string(),
                "FABRIC_E_INVALID_NAME_URI (-2147017794): The name 'fabric:/bad?x=1' is invalid: character '?' is not supported."
            );
```

`The name 'fabric:/bad?x=1' is invalid: character '?' is not supported.` This is the last error message and is very useful for debugging, that is never exposed in mssf before.

### DIFF
--- a/crates/libs/core/src/error/mod.rs
+++ b/crates/libs/core/src/error/mod.rs
@@ -139,14 +139,16 @@ impl From<core::num::TryFromIntError> for Error {
 // Get last error message from current thread from SF.
 // Call this after an SF API call failure.
 fn get_last_error_message() -> Option<WString> {
-    // The call returns error only when input result ptr is null:
+    // The call returns error only when input COM holder ptr is null:
     // https://github.com/microsoft/service-fabric/blob/ddfed33de371857f8fdb92287a8e0497297c8ccf/src/prod/src/retail/native/FabricCommon/FabricCommon.cpp#L233
-    let msg = crate::api::API_TABLE
-        .fabric_get_last_error_message()
-        .expect("failed to get error message");
-    // If message is not set, the returned string should be empty.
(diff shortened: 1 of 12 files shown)


crates/libs/core/src/error/mod.rs
```diff
@@ -18,69 +19,94 @@ pub type Result<T> = core::result::Result<T, Error>;
...
     fn from(value: Error) -> Self {
-        value.0
+        value.code
     }
 }
 
 impl From<crate::WinError> for Error {
     fn from(error: crate::WinError) -> Self {
-        Self(error.into())
+        Self::from_thread(error.code())
```
qbradley: It may not be a good idea to call Self::from_thread in the From implementation for Error. This hides a thread-local lookup inside an ".into()" call, or even hidden inside a ? which can call into() internally.

The contract for from_thread is that it is only valid immediately after an API call (there can not be any other intervening API calls or the value will not be valid).

qbradley (approved):

crates/libs/core/src/error/mod.rs
youyuanwu: Good catch. This is unintended. Changed to from_hresult.

PR #304, opened 2026-06-04
Proposal: gRPC xDS over Service Fabric naming resolve

youyuanwu:
See the doc.

### DIFF
--- a/docs/proposal/GrpcXds.md
+++ b/docs/proposal/GrpcXds.md
@@ -6,6 +6,44 @@ Date: 2026-06-04
 
 Owners: mssf maintainers
 
+## Why pursue this?
+
+**The value is client simplicity.** With `tonic-xds`, reaching an
+SF service is a vanilla xDS channel:
+
+```rust
+let channel = XdsChannelBuilder::with_config(
+    XdsChannelConfig::default()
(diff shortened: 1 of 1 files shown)


OscarTHZhang (commented):
So Option B is more like an Envoy deployed specifically to Service Fabric cluster?

docs/proposal/GrpcXds.md
```diff
@@ -0,0 +1,578 @@
...
+
+A Rust library that implements gRPC-xDS-the-client locally:
+parses `xds:///fabric/...` URIs, owns a `FabricClient` and
+naming plumbing, and feeds endpoints through gRPC's LB policy
+machinery.
+
+Pros: no extra process; pure in-process call.
+
+Cons: requires a Rust gRPC xDS client implementation. None
+exists in `tonic` today. Building one is a large undertaking
```
youyuanwu: Use tonic-xds

youyuanwu:
> So Option B is more like an Envoy deployed specifically to Service Fabric cluster?

Option A and B are similar to envoy sidecar. But A is per process and B is per vm/node.

chenwei75 (approved):

PR #308, opened 2026-06-19
[util] Split health producer into separate cluster/node and application loops

youyuanwu:
The application traversal (services, partitions, replicas) can be slow on large clusters and was delaying cluster/node health reporting. Run the two as independent loops concurrently so neither blocks the other. Track iterations separately via get_app_iteration and get_node_iteration.

Changed the health entity producer to emit loop completion events. Consumers should be responsible to collect stats.

### DIFF
--- a/crates/libs/util/src/monitoring/producer.rs
+++ b/crates/libs/util/src/monitoring/producer.rs
@@ -3,7 +3,10 @@
 // Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
 // ------------------------------------------------------------
 
-use crate::monitoring::{HealthEntity, NodeHealthEntity, entities::ClusterHealthEntity};
+use crate::monitoring::{
+    NodeHealthEntity, ProducerEvent,
+    entities::{ClusterHealthEntity, LoopKind},
+};
 use ::tokio::sync::mpsc;
 use mssf_core::{
     client::FabricClient,
@@ -21,11 +24,7 @@ use std::time::Duration;
 pub struct HealthDataProducer {
     fc: FabricClient,
     interval: Duration,
-    sender: mpsc::UnboundedSender<HealthEntity>,
-    /// Number of completed cluster + node loop iterations.
-    node_iteration: std::sync::atomic::AtomicU64,
-    /// Number of completed application loop iterations.
-    app_iteration: std::sync::atomic::AtomicU64,
+    sender: mpsc::UnboundedSender<ProducerEvent>,
 }
 
 /// Default timeout for FabricClient operations.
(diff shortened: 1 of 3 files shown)


crates/libs/util/src/monitoring/producer.rs
```diff
@@ -67,6 +77,20 @@ impl HealthDataProducer {
...
+    ///
+    /// This traversal can be slow on large clusters, so it runs in its own
+    /// loop independent of cluster/node reporting.
+    pub(crate) async fn run_once_applications(
+        &self,
+        token: BoxedCancelToken,
+    ) -> Result<(), Action> {
         // Get application information.
         if let Ok(apps) = self.get_all_applications(token.clone()).await {
             for app in apps {
```
jrahman, line 92: Is there any reason this has to be a serial loop vs. spawning multiple futures and collecting them? I'd expect most of the work here to be network calls to SF which can run concurrently.

crates/libs/util/src/monitoring/producer.rs
```diff
@@ -138,23 +162,50 @@ impl HealthDataProducer {
...
-        self.iteration
+        self.app_iteration
             .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
         Ok(())
     }
 
-    /// Run a loop to produce health data.
+    /// Run both the cluster/node loop and the application loop concurrently.
+    ///
+    /// The two loops are independent: a slow application traversal will not
```
jrahman, line 167: Technically doesn't even need two tasks, can create separate futures for each collection and then use something like FuturesUnordered or a JoinSet to poll them to completion.

crates/libs/util/src/monitoring/producer.rs
```diff
@@ -165,23 +216,34 @@ impl HealthDataProducer {
...
+    /// faster cluster/node loop has also produced data.
+    pub fn get_app_iteration(&self) -> u64 {
+        self.app_iteration
+            .load(std::sync::atomic::Ordering::Relaxed)
     }
 
-    pub fn get_iteration(&self) -> u64 {
-        self.iteration.load(std::sync::atomic::Ordering::Relaxed)
+    /// Number of completed cluster/node loop iterations.
+    pub fn get_node_iteration(&self) -> u64 {
```
jrahman: Are these only used for tests?

crates/libs/util/src/monitoring/producer.rs
```diff
@@ -138,23 +162,50 @@ impl HealthDataProducer {
...
+        tokio::join!(
+            self.run_cluster_node_loop(token.clone()),
+            self.run_application_loop(token.clone()),
+        );
+        tracing::info!("Health data producer loops exited.");
+    }
+
+    /// Run a loop to produce cluster and node health data.
+    pub async fn run_cluster_node_loop(&self, token: BoxedCancelToken) {
+        self.run_interval_loop(token, "cluster/node", |token| {
```
jrahman, line 179: Do we need both of these functions, or just inline the contents into the `tokio::join!()` above?

crates/libs/util/src/monitoring/producer.rs
```diff
@@ -67,6 +77,20 @@ impl HealthDataProducer {
                 }
             }
         }
+        self.node_iteration
```
jrahman: Only used for tests?

crates/libs/util/src/monitoring/producer.rs
youyuanwu: Spawning more futures could overload SF. We have thousands of partition+replica entities in the cluster, so parallelization needs to have some limit control. 
There are other health apis that might support bulk queries, or pagination etc, which can push the parallelization to the SF server side and I will explore them separately.

crates/libs/util/src/monitoring/producer.rs
youyuanwu: join! macro combines 2 futures on the same task. We don't have dependency of FuturesUnordered. JoinSet runs on multiple tasks and is an overkill.

crates/libs/util/src/monitoring/producer.rs
youyuanwu: Removed. I added producer event to indicate loop finished. Consumers can collect the stats based on the events.

crates/libs/util/src/monitoring/producer.rs
youyuanwu: inlined.

rsajwani (approved):

PR #311, opened 2026-06-30
Bulk implementation for Self-ReconfiguringService in mssf_core and sample app bring-up

OscarTHZhang:
Following the existing patterns for stateful and stateless core bindings, implement the core bindings for self-reconfiguring service.

Implement a dummy self-reconfiguring service that does a dummy leader election on the smallest instance id. Deployed and validated manually in a local SF Onebox with self-reconfiguring service feature switch enabled.

### DIFF
--- a/.devcontainer/onebox/ClusterManifest.SingleMachineFSS.xml
+++ b/.devcontainer/onebox/ClusterManifest.SingleMachineFSS.xml
@@ -130,6 +130,7 @@
     </Section>
     <Section Name="Management">
         <Parameter Name="ImageStoreConnectionString" Value="fabric:ImageStore" />
+        <Parameter Name="EnableSelfReconfiguringServices" Value="True" />
     </Section>
     <Section Name="NamingService">
       <Parameter Name="PartitionCount" Value="1" />
--- a/crates/libs/core/src/runtime/self_reconfiguring_traits.rs
+++ b/crates/libs/core/src/runtime/self_reconfiguring_traits.rs
@@ -32,22 +32,17 @@ pub trait ISelfReconfiguringServiceFactory: Send + Sync + 'static {
     ) -> crate::Result<Box<dyn ISelfReconfiguringServiceInstance>>;
 }
 
-/// Defines behavior that governs the lifecycle and configuration participation of
-/// a self-reconfiguring service instance, such as startup, configuration changes,
-/// and shutdown.
-///
-/// Unlike a stateless instance, a self-reconfiguring instance additionally
-/// receives configuration requests and configuration-change requests from Service
-/// Fabric and reports its resulting configuration through its partition.
+/// Safe abstraction over the `IFabricSelfReconfiguringServiceInstance` COM
+/// interface that an author implements. Each method corresponds to a method on
+/// that COM interface; `open` and `close` are asynchronous (COM Begin/End), the
+/// rest are synchronous.
 #[async_trait::async_trait]
 pub trait ISelfReconfiguringServiceInstance: Send + Sync + 'static {
-    /// Opens an initialized service instance so that it can be contacted by
-    /// clients. The returned string is the address of this service instance,
-    /// which is associated with the service name via Service Fabric naming and
-    /// returned to clients that resolve the service.
+    /// Opens an initialized service instance. The returned string is the address
+    /// of this service instance, which is associated with the service name via
+    /// Service Fabric naming and returned to clients that resolve the service.
     ///
-    /// The `open_mode` indicates whether the instance is being opened as new or
-    /// existing.
+    /// `open_mode` is the open mode supplied by Service Fabric.
     async fn open(
         &self,
         partition: Arc<dyn ISelfReconfiguringServicePartition>,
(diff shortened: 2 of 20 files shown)


crates/samples/dummy-self-reconfig/manifests/DummySelfReconfigServicePackage/ServiceManifest.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
youyuanwu, line 1: U can add the Service in the reflection test app, and serve it in the same process. We can write e2e tests via reflection app grpc server later.

.devcontainer/onebox/ClusterManifest.SingleMachineFSS.xml
```diff
@@ -130,6 +130,7 @@
     </Section>
     <Section Name="Management">
         <Parameter Name="ImageStoreConnectionString" Value="fabric:ImageStore" />
+        <Parameter Name="EnableSelfReconfiguringServices" Value="true" />
```
youyuanwu: Enable this for windows as well:
https://github.com/Azure/service-fabric-rs/blob/main/onebox/windows/ClusterManifestTemplate.json

crates/samples/dummy-self-reconfig/manifests/DummySelfReconfigServicePackage/ServiceManifest.xml
youyuanwu: Add write a reflection test to query app is available using fabricclient. Can be a separate PR

crates/samples/dummy-self-reconfig/src/instance.rs
```diff
@@ -0,0 +1,251 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License (MIT). See License.txt in the repo root for license information.
+// ------------------------------------------------------------
+
+//! Self-reconfiguring service instance.
+//!
+//! The "business logic" is modeled on the internal Service Fabric
+//! `SelfReconfigTestService` C++ sample: the instance with the lowest instance
+//! id is elected leader, and only the leader reports the configuration back to
```
jrahman: > only the leader reports the configuration back to
Service Fabric

Is that the official contract for Self-Reconfiguring services?

crates/libs/core/src/runtime/self_reconfiguring_traits.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// `open_mode` is the open mode supplied by Service Fabric.
+    async fn open(
+        &self,
+        partition: Arc<dyn ISelfReconfiguringServicePartition>,
+        open_mode: SelfReconfiguringOpenMode,
+        cancellation_token: BoxedCancelToken,
+    ) -> crate::Result<WString>;
+
+    /// Synchronous notification of a configuration request from Service Fabric
+    /// (COM `RequestConfiguration`).
```
jrahman: Comment here is a bit sparse, what is the expected behavior/semantics for this callback implementation?

crates/libs/core/src/runtime/self_reconfiguring_traits.rs
```diff
@@ -0,0 +1,114 @@
...
+        health_info: &crate::types::HealthInformation,
+    ) -> crate::Result<()>;
+
+    /// Reports current partition health.
+    fn report_partition_health(
+        &self,
+        health_info: &crate::types::HealthInformation,
+    ) -> crate::Result<()>;
+
+    /// Reports configuration through the partition (COM `ReportConfiguration`).
```
jrahman: Sparse comments here as well, more semantic details on what SF expects from the contents of the Configuration report would be helpful.

.devcontainer/onebox/ClusterManifest.SingleMachineFSS.xml
OscarTHZhang: Added

crates/samples/dummy-self-reconfig/src/instance.rs
OscarTHZhang: It is, but because this is currently an internal feature, I decided not to expose that in the public documentation for Rust SDK. The relevant method documentation is also marked as TBD for now.

crates/libs/core/src/runtime/self_reconfiguring_traits.rs
OscarTHZhang: Marked the public documentation explicitly to TBD for now as this is an internal feature.

crates/samples/dummy-self-reconfig/manifests/DummySelfReconfigServicePackage/ServiceManifest.xml
OscarTHZhang: Booked an item #314

crates/libs/core/src/runtime/self_reconfiguring_traits.rs
OscarTHZhang: Explicitly marked the doc as TBD

crates/libs/core/src/types/runtime/self_reconfiguring.rs
```diff
@@ -0,0 +1,606 @@
...
+    pub instances: Vec<InstanceChangeRequest>,
+}
+
+impl From<&FABRIC_SELF_RECONFIGURING_CONFIGURATION_CHANGE_REQUEST>
+    for SelfReconfiguringConfigurationChangeRequest
+{
+    fn from(r: &FABRIC_SELF_RECONFIGURING_CONFIGURATION_CHANGE_REQUEST) -> Self {
+        let mut instances = Vec::new();
+        if !r.Instances.is_null() {
+            let list = unsafe { &*r.Instances };
```
youyuanwu: THere should be some helpers to convert arrays. U do not need for loops.

youyuanwu (approved):

crates/libs/core/src/types/runtime/self_reconfiguring.rs
OscarTHZhang: applied

PR #312, opened 2026-07-02
Add support for application upgrade info

belee325:
This PR adds support for getting information about inflight UD walks for applications on an SF cluster


crates/libs/util/src/monitoring/upgrade_producer.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /// Run once to produce application upgrade data.
+    ///
+    /// Enumerates all applications and, for each one that is actively upgrading,
+    /// emits an [`UpgradeProducerEvent::Upgrade`]. Applications that are not
+    /// upgrading are skipped. An [`UpgradeProducerEvent::IterationComplete`]
+    /// marker is emitted once all applications have been processed.
+    pub(crate) async fn run_once(&self, token: BoxedCancelToken) -> Result<(), Action> {
+        if let Ok(apps) = self.get_all_applications(token.clone()).await {
```
youyuanwu: Should this be merged with the existing health producer events? Withe a config opiton to emit it?

belee325: ah yeah, we already query for deployed apps in that one. 

will address in next revision

belee325:
@microsoft-github-policy-service agree company="Microsoft"

rsajwani (approved):

crates/libs/core/src/types/client/application_upgrade.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// Returning failed as true, since a failed state is actionable.
+    /// The completed / invalid states return false.
+    pub fn is_active(&self) -> bool {
+        matches!(
+            self,
+            Self::RollingForwardPending
+                | Self::RollingForwardInProgress
+                | Self::RollingBackPending
+                | Self::RollingBackInProgress
+                | Self::Failed
```
youyuanwu, line 48: Failed is active?

youyuanwu (approved):

crates/libs/core/src/types/client/application_upgrade.rs
belee325: Maybe should've called it is_actionable or something. Wanted to filter out terminal states, but wanted to include failed states so on-call can intervene. 

Will address in follow up

PR #322, opened 2026-08-06
[Prototype] Add mssf-net: SF naming to gRPC xDS (ADS) mapping

youyuanwu:
> ## ⚠️ Prototype — narrow slice, not production-ready
>
> This is an **experiment**, not a finished feature. It implements only
> **Phase 1 + Phase 2** of the `GrpcXds.md` proposal (merged in #321) — enough to
> prove the SF→xDS concept end-to-end — and deliberately stops there.
>
> | | |
> |---|---|
> | **Implemented** | One SF service per ADS server; stateful **singleton partition, primary only**; live updates via SF change notifications; verified against a real onebox failover. |
> | **Not implemented** | Multiple services per server · partition-key routing · secondaries / read-from-secondary · TLS/mTLS · RDS · delta xDS · federation (`xdstp://`) · load reporting (LRS/ORCA) · the agent/relay deployment shape. |
>
> Other caveats worth knowing before reviewing:
>
> - Depends on the **pre-release `tonic-xds` 0.1.0-alpha.2**, so the crate is
>   `publish = false` and its error *text* is not treated as a stable contract.
> - The crate is marked **experimental with no stable API guarantee**.
> - **Nothing existing changes behaviour.** The incumbent `mssf_util::tonic` path
>   is untouched; the only edit to an existing crate makes four private test
>   helpers public so two tests can share one scenario definition.
>
> Every exclusion above is documented in `docs/design/XdsNamingDesign.md` and in
> the deferred-work list — none of it is an accidental gap. Reviewing this as
> "is the shape right and is the concept proven?" will be more useful than
> "is this feature-complete?", because it is not.

## What

Adds `mssf-net` (`crates/libs/net`), a library that maps Service Fabric naming onto gRPC xDS
and serves it over a State-of-the-World ADS server, so a **stock, unmodified** xDS gRPC client
can reach an SF service with **no SF-specific client code**.

It is a **library**, not the binary agent the proposal sketched — a future `mssf-xds-agent` /
`mssf-xds-discovery` process can be built on top without changing this layer.

What a caller writes:

```rust
let target  = XdsUri::parse("xds:///fabric:/MyApp/MyService")?;
let channel = XdsChannelBuilder::new(XdsChannelConfig::new(target)).build_grpc_channel()?;
let mut client = GreeterClient::new(channel);
```

No `FabricClient`, no `ServicePartitionResolver`, no endpoint parsing, no replica selection, no
reconnect loop — and **no dependency on `mssf-net`**; the only new client dependency is
`tonic-xds`. Because it is ordinary xDS, Go/Java/C++/Node clients reach the same service with
their own standard xDS channels, which is the reason to prefer this over the Rust-only path.

Hosting the mapping:

```rust
let mapping = XdsMapping::for_service_uri("fabric:/MyApp/MyService", host_port_interpreter())?;
let source  = FabricEndpointSource::new(&mapping, conn_strings, timeout).await?;
let server  = AdsService::new(mapping, source.clone()).serve_on_ephemeral_loopback().await?;
```

## Proof it works

Against a real local onebox, driving a stock `tonic-xds` client through this crate to the
`reflection` sample, using only metadata-free RPCs:

```
steady-state primary      replica_before=134305279242932538
   ... restart_replica failover ...
recovered on new primary  replica_after=134305279255905545
```

Same client, same ADS server, neither restarted.

| Check | Result |
|---|---|
| Unit tests | 43 pass |
| Cluster-free e2e (real `tonic-xds` client) | 7 pass |
| README examples type-check | pass |
| Live onebox `xds_failover` | pass |
| Live onebox `tonic_failover` (regression) | pass |
| `clippy --all-targets -- -D warnings`, `fmt`, rustdoc `-D warnings` | clean |

## Design notes

- **`EndpointSource` trait is the SF seam.** With SF behind a trait, the whole mapping and
  serving stack is exercised end-to-end against a real xDS client with **no cluster, no SF
  runtime, no `FabricClient`**. The cluster-backed CI jobs already cover execution via
  `cargo test --all`; the one added CI step is in `build-azl3`, which has no SF installed and
  otherwise runs no tests, making it both new coverage and the strongest no-SF-runtime proof.
- **The SF URI is the xDS resource name.** `xds:///fabric:/App/Service` — so the name a client
  writes *is* the service it wants, with no alias table to keep in sync. Works because the
  virtual host uses `domains: ["*"]`; a literal domain match would have to reproduce the URI.
- **LDS with an inline `RouteConfiguration`; RDS is not needed.** The proposal assumed
  LDS→RDS→CDS→EDS; inlining the route config removes a resource type with no loss of function.
- **Three endpoint states, not two.** `Primary` / `NoPrimary` / `NotFound`. The split is
  load-bearing: `NoPrimary` publishes an empty assignment (client reports "no ready
  endpoints"), while `NotFound` withholds the Listener — and since Listeners are
  all-resources-required in SotW, that omission is a *deletion*. `ServicePartitionResolver`
  collapses both into `FABRIC_E_SERVICE_OFFLINE`, so `classify_resolve_error` classifies
  explicitly and maps only known does-not-exist codes to the permanent state.
- **The COM callback does almost nothing.** SF's notification callback is synchronous on a COM
  thread, so it only does a non-blocking `send_replace`; a Tokio task interprets addresses.
  `tokio::sync::watch` gives latest-value-wins coalescing structurally.
- **Addresses are opaque.** SF endpoint addresses are service-defined strings — the reflection
  sample publishes a URL with query parameters, not `host:port` — so interpretation is a
  caller-supplied function. It takes `&str`, not `&ResolvedServicePartition`, because
  notifications carry no RSP.
- **Shutdown is a `CancellationToken`**, matching this repo's existing convention. One token
  both ends the open ADS streams and triggers graceful stop — necessary because an ADS stream
  is long-lived and `tonic` waits for connections to drain, so stopping only the accept loop
  hangs. `with_cancellation(..)` accepts a caller's token so the server can be tied to an SF
  replica's `close`.

Full rationale: `docs/design/XdsNamingDesign.md`.

## Notes for reviewers

- `crates/libs/net` needs an explicit `members` entry — `crates/libs/*` is not globbed.
  The crate depends on `mssf-util` *without* its `tonic` feature, so the successor cannot link
  the incumbent.
- Three bugs were found and fixed in review, each with a regression test where one is
  meaningful: ACK discrimination used `version_info` instead of `response_nonce` (a
  reconnecting client replays its retained version with an empty nonce, so the server would
  have ignored every request on the new stream and stalled discovery permanently);
  `NotFound`→`Primary` never re-sent LDS/CDS; and a startup seed/notification race.
- `tests/readme_check.rs` type-checks the README examples so they cannot drift. Writing it
  caught `ServerHandle::shutdown` returning `Box<dyn Error + Send + Sync>`, which does not
  convert via `?` into the `Box<dyn Error>` most callers return — now a `ShutdownError` enum.
- The `biased;` on the stream's `select!` is for deterministic shutdown, not liveness — an
  unbiased select still wins with probability 1, so no black-box test distinguishes them. The
  accompanying test is labelled a smoke test rather than a guard it cannot provide.
- If `tonic_failover` fails with `Unimplemented` locally, check the *deployed* app package —
  a months-old binary predating the `Write` RPC cost me a debugging cycle.


chenwei75:
Some minor comments from copilot, 
Subscription changes carrying a nonce are silently dropped. is_subscription_request returns false for any request with a non-empty response_nonce, and the handler continues. But a real SotW client that changes its resource set on an existing stream sends the new resource_names alongside the last nonce. The subscription map is updated, but no response is generated for it — so the client waits for a resource it will only receive on the next endpoint change. Worth either handling "ACK whose resource_names differ from the recorded set" as a subscription, or explicitly documenting the assumption that clients never re-subscribe mid-stream.
bootstrap_json interpolates node_id without escaping. A node id containing " or \ produces invalid JSON. One-line fix: serialize the field or assert on the character set.
A NACK is logged and then forgotten. tracing::warn! and continue means a persistently rejected resource never gets retried or resent, and there's no metric or surfaced signal. Fine for a prototype, but it's the failure mode that will be hardest to diagnose in the field — worth a note in the deferred list.
cluster_name() = format!("{}-primary", xds_name) is collision-prone once multiple services per server land: fabric:/A/Svc-primary and fabric:/A/Svc would both derive the same cluster. Since the multi-service map is already sketched in the design doc, a reserved separator or explicit uniqueness check would cost little now.
The notification-interpreting Tokio task isn't explicitly stopped. FabricEndpointSource::shutdown unregisters the filter and drops the client, and the task exits when raw_tx drops with it — but that's an implicit dependency on the callback closure being the sole remaining sender. Holding a JoinHandle (or a child CancellationToken) would make the teardown ordering explicit rather than inferred.
Minor: crates/samples/reflection/Cargo.toml still has no trailing newline; build.rs compiles the test-only testsvc.proto on every build including non-test ones.

youyuanwu:
Thanks — I went through these against the code rather than taking them at face value. **Four are valid and fixed, one I believe is incorrect as stated, one is accepted as a documented cost.** Pushed in 06db44e.

---

**1. Subscription changes carrying a nonce are silently dropped — VALID, fixed.**

This is the most important one, and it's real rather than theoretical. I checked `xds-client` 0.1.0-alpha.2: `send_request` (`worker.rs:~925`) always populates both `version_info` and `response_nonce` from the per-type state, and it's called on both `Watch` and `Unwatch`. So once a type has received one response, any later resource-set change on that stream carries a nonce and my discriminator classified it as an ACK.

It's latent *today* only because a single-mapping server never changes a client's resource set — the LDS→CDS→EDS cascade subscribes each type once, and the first request per type has an empty nonce. But it would fire immediately once multi-service lands, which the design doc already sketches. Exactly the kind of thing that would have been miserable to debug later.

Fixed by comparing against the previously recorded set: a request whose `resource_names` differ is treated as a subscription regardless of the nonce. Unit test covers both directions (same set + nonce → skip; changed set + nonce → answer).

**2. `bootstrap_json` doesn't escape `node_id` — VALID, fixed.**

Straightforwardly broken for any id containing `"` or `\`. Now escaped per RFC 8259 including control characters, and the test asserts the escaped output still parses as a `BootstrapConfig` rather than just eyeballing the string. (`ads_addr` is a `SocketAddr`, so its `Display` is always JSON-safe — left alone.)

**3. The notification-interpreting task isn't explicitly stopped — VALID, fixed.**

Your reading is exactly right: it *did* terminate, but only because the callback closure owned by the `FabricClient` held the last `watch` sender, so dropping the client ended the loop. Correct by inference, which is not a property I want teardown to rely on.

It now owns a `CancellationToken` and its `JoinHandle` is held in the releasable state, so `shutdown` cancels and **awaits** it — and does so *before* unregistering the filter, so nothing can publish mid-teardown.

**4. Trailing newline on `crates/samples/reflection/Cargo.toml` — VALID, fixed.** My edit dropped it.

---

**5. `cluster_name()` collision — I don't think this one holds.**

`format!("{}-primary", xds_name)` appends a constant suffix, which is injective: `f(a) == f(b)` implies `a == b`. So two distinct xDS names cannot derive the same cluster name. Taking the example literally, `fabric:/A/Svc-primary` derives `fabric:/A/Svc-primary-primary`, while `fabric:/A/Svc` derives `fabric:/A/Svc-primary` — different strings.

That said, there's a real adjacent risk worth capturing, so I've added it to the design doc rather than dismissing the point: a service named `fabric:/X-primary` has a *Listener* named `fabric:/X-primary`, and a service named `fabric:/X` has a *Cluster* named `fabric:/X-primary`. Those are different xDS resource types with separate namespaces (my `resources_for` matches on `type_url` first), so it's fine today — but a future multi-service registry keyed by bare name instead of `(type_url, name)` would conflate them. The design doc now says: when multi-service lands, validate uniqueness of xDS names and key the registry by type + name.

Happy to be corrected if you had a different collision in mind.

---

**6. NACK-and-forget, and `build.rs` on every build — VALID, documented rather than fixed.**

Both agreed, and both now in the deferred list with reasoning:

- **NACK handling.** Logging and not advancing state is correct, but a persistently rejected resource is never re-sent and raises no metric. I called out *why* it's the nastiest one to diagnose: the client silently has no route while the server looks healthy. A real deployment wants a counter plus a surfaced error on repeated NACKs for the same resource.
- **`protoc` as a build requirement.** True — a library-only consumer pays to compile a test fixture. Noted, with the fix (feature-gate the stand-in service or move it to a separate test crate). Left as-is for now since CI already installs `protoc` for the other crates, but it's a genuine wart.

---

**Verification:** 45 unit + 7 cluster-free e2e + README compile-guard pass; clippy `-D warnings`, fmt, rustdoc clean.

⚠️ **One caveat:** my onebox is currently down, so I could not re-run the live `xds_failover` test after the `shutdown`-ordering change in (3). The change is low-risk — cancel + await a `JoinHandle` ahead of the existing teardown — and the cluster-free suite covers the ADS changes in (1) and (2), but the live path hasn't been re-verified since that edit. I'll re-run it before this merges.

youyuanwu:
**Live verification done** — the caveat in my previous comment is now cleared.

Onebox back up, `fabric:/ReflectionApp` deployed and Ready (deployed package verified current, i.e. it carries both `Write` and `GetReplicas`, so no stale-binary confusion this time).

**Live tests, after the `shutdown`-ordering change:**

```
xds_failover:   steady-state primary   replica_before=134309594827570420
                ... restart_replica failover ...
                recovered on new primary  replica_after=134309594931762825   ✅ ok

tonic_failover: ✅ ok  (incumbent path — no regression from the helper extraction)
```

**Full workspace suite, `cargo test --all` against the live cluster: 222 tests, 0 failures**, across all 35 test binaries. Nothing in the repo regressed.

So the two fixes that could plausibly have disturbed runtime behaviour are both confirmed on a real cluster:

- the ADS resource-set/nonce change (finding 1) — exercised by every live xDS resolution in the run;
- the explicit notification-task teardown (finding 3) — exercised by `xds_failover`'s teardown path, which now cancels and awaits the task before unregistering the filter.

Ready for another look whenever you have time. Still very much a prototype in scope — see the callout at the top of the description for what is and isn't implemented.

chenwei75 (approved):

PR #324, opened 2026-08-13
Made changes to support "DeployedApplication / DeployedServicePackage" entities in the QueryClient. Also added missing fields and additional utility functions to some entities.

That-Canadian:
Made changes to support "DeployedApplication / DeployedServicePackage" entities in the QueryClient. Also added missing fields and additional utility functions to some entities.

Also bumped mssf-core from 0.8.3 -> 0.8.4

QueryClient
- get_deployed_application_list / get_deployed_service_package_list

NodeQueryResultItem:
- Added missing 'NodeStatus' and 'HealthState' mappings

'ServiceQueryResultItem':
- get_service_type_name
- get_service_manifest_version
- get_service_status

'ServicePartitionQueryResultItem':
- get_partition_information
- get_partition_status

'ServiceReplicaQueryResultItem':
- get_replica_status
- get_node_name

'ReplicaHealth'
- get_health_events
- get_replica_or_instance_id
- get_partition_id

### DIFF
--- a/crates/libs/core/src/types/client/node.rs
+++ b/crates/libs/core/src/types/client/node.rs
@@ -4,8 +4,10 @@
 // ------------------------------------------------------------
 
 use crate::mem::{BoxPool, GetRawWithBoxPool};
+use crate::types::HealthState;
 use crate::{WString, types::Uri};
 use bitflags::bitflags;
+use mssf_com::FabricTypes::FABRIC_QUERY_NODE_STATUS;
 use mssf_com::{
     FabricClient::IFabricGetNodeListResult2,
     FabricTypes::{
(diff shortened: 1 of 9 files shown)


crates/libs/core/src/types/client/node.rs
```diff
@@ -145,7 +147,9 @@ impl From<&FABRIC_NODE_QUERY_RESULT_ITEM> for NodeQueryResultItem {
             node_type: WString::from(raw.NodeType),
             code_version: WString::from(raw.CodeVersion),
             config_version: WString::from(raw.ConfigVersion),
+            status: value.NodeStatus.into(),
```
youyuanwu, line 150: Keep the name the same as bindings: aggregated_health_state

youyuanwu, line 150: nodes_status

youyuanwu (approved):

PR #325, opened 2026-08-19
Support VeryHigh replica move cost

wujysh:
## Summary

- expose `MoveCost::VeryHigh` from `mssf-core`
- map it to and from the native `FABRIC_MOVE_COST_VERYHIGH` value

## Validation

- `cargo +ms-prod-1.95 fmt --check`
- `cargo +ms-prod-1.95 check -p mssf-core`
- `cargo +ms-prod-1.95 clippy -p mssf-core --all-targets -- -D warnings`

The repository-pinned Rust 1.97.1 toolchain was unavailable in the local environment; GitHub CI can provide exact-version validation.


wujysh:
@youyuanwu If there is special reason this cannot be wired up, please let me know. From mssf_com, it is available.

youyuanwu (approved):

youyuanwu:
Originally it was omitted due to the SF idl/header version too low and no constants generated. Now it is available.

PR #328, opened 2026-08-27
Change usage of 'FILETIME' in mssf-core to 'SystemTime'.

That-Canadian:
Change usage of 'FILETIME' in mssf-core to 'SystemTime' and created a helper function to handle the conversion, based on the following Microsoft doc: [FILETIME Structure](https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-filetime)

Also bump ms…sf-core from 0.8.5 -> 0.8.6

### DIFF
--- a/crates/libs/core/src/time.rs
+++ b/crates/libs/core/src/time.rs
@@ -19,7 +19,7 @@ const FILETIME_UNIX_EPOCH_OFFSET_TICKS: u64 = 116_444_736_000_000_000;
 /// 100-nanosecond precision. Both epochs denote UTC instants, although
 /// [`SystemTime`] itself has no timezone. Returns [`None`] when the result is
 /// outside the range supported by the current platform.
-pub(crate) fn filetime_to_system_time(filetime: FILETIME) -> Option<SystemTime> {
+pub(crate) fn try_filetime_to_system_time(filetime: FILETIME) -> Option<SystemTime> {
     // FILETIME represents one unsigned 64-bit tick count as separate high and
     // low 32-bit words. Shifting the high word restores bits 32 through 63.
     let ticks = (u64::from(filetime.dwHighDateTime) << 32) | u64::from(filetime.dwLowDateTime);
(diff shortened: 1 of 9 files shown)


crates/libs/core/src/time.rs
```diff
@@ -0,0 +1,46 @@
...
+/// and Rust's [`UNIX_EPOCH`] (`1970-01-01 00:00:00 UTC`). The two epochs are
+/// 134,774 days, or 11,644,473,600 seconds, apart:
+/// `11,644,473,600 seconds * 10,000,000 ticks/second`.
+const FILETIME_UNIX_EPOCH_OFFSET_TICKS: u64 = 116_444_736_000_000_000;
+
+/// Converts a Windows [`FILETIME`] to a [`SystemTime`], preserving its
+/// 100-nanosecond precision. Both epochs denote UTC instants, although
+/// [`SystemTime`] itself has no timezone. Returns [`None`] when the result is
+/// outside the range supported by the current platform.
+pub(crate) fn filetime_to_system_time(filetime: FILETIME) -> Option<SystemTime> {
```
youyuanwu: try_filetime_to_system_time

youyuanwu: Add another function for filetime_to_system_time that fills the default value when conversion fails

crates/libs/core/src/types/client/node.rs
```diff
@@ -341,17 +344,19 @@ mod tests {
...
                 dwLowDateTime: 100,
                 dwHighDateTime: 200,
-            }
+            })
+            .unwrap()
         );
         assert_eq!(
             item.node_down_at,
-            FILETIME {
+            filetime_to_system_time(FILETIME {
```
youyuanwu: Do we have sanity test that the conversion result is reasonable with some hardcoded well known values?

crates/libs/core/src/time.rs
That-Canadian: Done

crates/libs/core/src/types/client/node.rs
That-Canadian: Added some hardcoded tests, one for UNIX Epoch and another for an arbitrary time (May 12 2026 10:29:43.004771)

youyuanwu (approved):

youyuanwu (approved):

PR #333, opened 2026-09-18
Update Windows bindings to crates.io 0.100

youyuanwu:
## Summary

- upgrade the published crates.io dependencies from `windows-bindgen` 0.66 to 0.100 and `windows-core` 0.62 to 0.100, then regenerate `mssf-com`
- keep the `windows` projection crate on its latest published 0.62.2 release
- adapt the safe wrappers and PAL compatibility layer to the new generated ID, URI, FILETIME, and COM signatures
- remove the obsolete PAL FILETIME shim and document follow-up investigation for adopting upstream cross-platform PCWSTR and WIN32_ERROR safely
- raise the workspace MSRV to Rust 1.95 and source the CI minimum-version matrix entry from Cargo metadata
- generate bindings from the published `mssf-metadata` crate instead of fetching the metadata repository through CMake

## Validation

- `cargo +1.95.0 check --all-targets --jobs 2`
- `cargo clippy --workspace --all-targets --jobs 2 -- -D warnings`
- `cargo fmt --all -- --check`
- `cargo test -p mssf-com -p mssf-pal -p mssf-core --lib --jobs 2 -- --skip test_fabric_client`
- CMake configure and `generate_rust` target on Windows
- full Windows CMake build and sample packaging

The cluster-dependent `test_fabric_client` was excluded because the local Service Fabric gateway was not running.


crates/tools/api/src/main.rs
```diff
@@ -2,53 +2,121 @@
...
+        "FabricRuntime",
+    ] {
+        let path = format!("crates/libs/com/src/Windows/ServiceFabric/{ns}/mod.rs");
+        emit_enum_const_aliases(&path);
+    }
+}
+
+/// Appends free module-level constant aliases for every scoped-enum associated
+/// constant in `path` (see call site for rationale).
+fn emit_enum_const_aliases(path: &str) {
```
youyuanwu: Do not reparse generated code. If it is breaking changes we need to fix down stream.

.gitignore
```diff
@@ -7,4 +7,9 @@ obj
 .reviews
 
 # cargo stuff
-target
\ No newline at end of file
+target
+
+# Local copy of the SF winmd used by tools_api during the windows-bindgen
+# experiment. Normally the winmd is fetched via CMake FetchContent into
+# build/_deps; this is a generated binary artifact and must not be committed.
```
youyuanwu: Update comment. cmake fetch content is no longer used.

crates/tools/api/src/main.rs
```diff
@@ -2,53 +2,121 @@
...
+    let workspace = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .parent()
+        .and_then(|path| path.parent())
+        .and_then(|path| path.parent())
+        .expect("tools_api must be located under crates/tools")
+        .to_path_buf();
+    env::set_current_dir(&workspace)
+        .unwrap_or_else(|e| panic!("change directory to {}: {e}", workspace.display()));
+
+    let old_namespace = workspace.join("crates/libs/com/src/Microsoft");
```
youyuanwu: Old code has been removed. No need to keep the migration code.

.gitignore
youyuanwu: Fixed in e273c69. Removed the obsolete tools_api/winmd ignore and its FetchContent comment.

crates/tools/api/src/main.rs
youyuanwu: Fixed in e273c69. Removed the old Microsoft namespace cleanup migration from tools_api.

crates/tools/api/src/main.rs
youyuanwu: Fixed in e273c69. Removed generated-code reparsing and the compatibility aliases; downstream callers now use the generated scoped enum associated constants directly.

wujysh (approved):

